PR #13, opened 2019-02-11
Merge in macros/tools from DLME

jcoyne:


peetucket:
circleci can't find rspec .. is it something to do with the fact that the rspec gem is in the `test` group?

jcoyne:
@peetucket it looks good to me.

PR #17, opened 2019-02-11
Write to ndjson rather than to the database

jcoyne:


lib/dlme_json_resource_writer.rb
```diff
@@ -4,47 +4,11 @@
 
 require_relative 'adjust_cardinality'
```
peetucket: does this still need to `require_relative` if they are in the same folder?

jcoyne: I don't know. That line was not part of this change.

jcoyne:
@peetucket I'm basing this on https://github.com/traject/traject/blob/master/lib/traject/json_writer.rb#L42-L49

PR #18, opened 2019-02-12
Set the exec bit on invoke.sh

jcoyne:


justinlittman:
Can you describe the reason for the change?

jcoyne:
@justinlittman this allows us to run invoke.sh outside of (while not changing it's ability to run within) the container primarily for debug purposes.  Where the container is pretty opaque about what is happening.

justinlittman:
Ahh, I missed the change to invoke.sh itself.

PR #20, opened 2019-02-12
Copy all the code into the container

jcoyne:


justinlittman:
When would we want copied code? I can imagine that there are 2 ways that we would want to run this: (1) with current, local code or (2) with master code. For 1, easier to link in code at runtime rather than rebuild container. For 2, set USE_GITHUB to true to pull latest at runtime.

Happy to discuss this more.

jcoyne:
@justinlittman without this the `lib/` directory doesn't make it into the container and the builds fail with:
```
Error loading configuration file config/mods_config.rb:3 LoadError:cannot load such file -- xml_reader
```

PR #30, opened 2019-02-13
Validate all outputs

### LINKED ISSUE #27: Add a schema validator
Ensure that all records conform to the schema.

jcoyne:
Fixes #27


lib/dlme_json_resource_writer.rb
```diff
@@ -1,14 +1,26 @@
...
   def serialize(context)
     attributes = context.output_hash.dup
     id = attributes.fetch('id').first
-    JSON.generate(AdjustCardinality.call(attributes)).unicode_normalize
+    json = AdjustCardinality.call(attributes)
+    errors = validate(json)
+    if errors.empty?
+      JSON.generate(json).unicode_normalize
+    else
+      context.logger.error("Row for #{id} is invalid")
```
justinlittman: I think that log then fail. Otherwise, validation errors will be missed.

lib/dlme_json_resource_writer.rb
```diff
@@ -1,14 +1,26 @@
 # frozen_string_literal: true
 
 require 'traject'
-
-require_relative 'adjust_cardinality'
+require 'adjust_cardinality'
+require 'dlme_json_schema'
 
 # Write the traject output to newline delmitited json
```
justinlittman: Typo.

lib/dlme_json_resource_writer.rb
justinlittman: Log JSON and errors?

lib/dlme_json_resource_writer.rb
jcoyne: That's not part of this PR.

justinlittman: But it offends my aesthetic sensibilities ...

lib/dlme_json_resource_writer.rb
jcoyne: There is no JSON at this point. Just a hash.

jcoyne: Why bother logging if we're going to fail? Can we just fail with a message?

justinlittman: Right, but could easily generate.

justinlittman: Works for me.

jcoyne:
@justinlittman changes made. Can you look again?

PR #35, opened 2019-02-14
Wrapper around Traject transform, providing ability to transform all …

### LINKED ISSUE #25: Support transforming all

justinlittman:
…and convenient configuration of Traject config files, settings.

closes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,7 @@ everything is update to date on the master branches):
 ```
 docker run --rm -v $(pwd)/output:/opt/traject/output \
                 suldlss/dlme-transform:latest \
-                -c config/mods_config.rb \
-                -s agg_provider=Stanford \
-                -s agg_data_provider=Stanford \
-                dlme-metadata/stanford/maps/data/kj751hs0595.mods
+                stanford/maps/data/kj751hs0595.mods
 ```
 
 The output will appear in STDOUT inside the container and be written to
@@ -59,18 +56,19 @@ Specify the Traject config file to use with the -c switch as shown below (e.g. `
 docker run --rm -e SKIP_FETCH_CONFIG=true \
                 -e SKIP_FETCH_DATA=true \
                 -v $(pwd)/../dlme-traject:/opt/traject/config \
-                -v $(pwd)/../dlme-metadata:/opt/traject/dlme-metadata \
+                -v $(pwd)/../dlme-metadata:/opt/traject/data \
                 -v $(pwd)/output:/opt/traject/output \
                 suldlss/dlme-transform:latest \
-                -c config/mods_config.rb \
-                -s agg_provider=Stanford \
-                -s agg_data_provider=Stanford \
-                dlme-metadata/stanford/maps/data/kj751hs0595.mods
+                stanford/maps/data/kj751hs0595.mods
 ```
 
-To process a directory of files, use a wildcard in the input filepath. For
-example, `dlme-metadata/stanford/maps/data/*.mods` instead of
-`dlme-metadata/stanford/maps/data/kj751hs0595.mods`.
+To process multiple files, specify a directory instead of a single file. For
+example, `stanford/maps` instead of `stanford/maps/data/kj751hs0595.mods`. To transform everything, specify nothing.
+
+```
+docker run --rm -v $(pwd)/output:/opt/traject/output \
+                suldlss/dlme-transform:latest
+```
 
 
 Sending output to S3 bucket:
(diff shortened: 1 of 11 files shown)


justinlittman:
Still need to write tests.

metadata_mapping.json
```diff
@@ -0,0 +1,13 @@
+{
+  "mods_config.rb": {
+    "paths": [
+      "stanford/maps"
+    ],
+    "extension": ".mods",
```
jcoyne: This seems a little clunky.  Could `paths` support wildcards? e.g. `stanford/maps/**/*.mods`?

metadata_mapping.json
```diff
@@ -0,0 +1,13 @@
+{
+  "mods_config.rb": {
```
jcoyne: I don't think this is a good top level key, because multiple collections could use the same transform file.  Perhaps this should be modeled on https://github.com/sul-dlss/dlme/blob/master/config/settings.yml#L42-L153 ?

metadata_mapping.json
justinlittman: The idea is to be able to figure out from a specified directory which config to use for it and how to find the files within it. Thus, for `stanford/maps/dir1`, would be able to determine that matches `stanford/maps` so should use `mods_config.rb` and that within `stanford/maps/dir1` (and descendants) should look for `.mods` files. I don't think that would be easy with wildcards but open to suggestions.

metadata_mapping.json
justinlittman: Good point. Will take that into account.

metadata_mapping.json
jcoyne: Can we do this?
```ruby
hash.each do |key, val|
  val['paths'].each do |glob|
    if  Dir.glob(val['path'].).any?
      #Do the stuff
    else 
       # NOP
    end
  end
end
```

justinlittman: I want to make sure we're not talking past each other, so let me back up. I'd like a user to be able to use `exe/transform` for all traject invocations, not just all. (So, yes, I expanded the scope of this ticket to simplify performing transforms.) Thus, a user should be able to:
1. Provide no filepath, in which case everything is transformed.
2. Provide a filepath that is a parent of multiple collections, in which case the appropriate files are selected for each collection and the appropriate traject config is used for each collection.
3. Provide a filepath for a collection, in which case the appropriate files are selected and the appropriate traject config is used.
4. Provide a filepath within a collection, in which case the appropriate subset of files is selected and appropriate traject config is used.
5. Provide a filepath for a single file within a collection, in which case the appropriate traject config is used.

More broadly, the user just needs to provide a filepath and `exe/transform` will determine which files and traject config to use.

The problem with using a wildcard for a path (as I see it), is that it makes it difficult to match the specified filepath with the config. For example, suppose the specified filepath is `stanford/maps/dir1`. If path is `stanford/maps` then can determine that it is a parent of `stanford/maps/dir1`. However, cannot determine that if path is `stanford/maps/**/*.mods`.

At least I think. Happy to pair on this.

README.md
```diff
@@ -79,9 +77,34 @@ docker run --rm -e S3_BUCKET=s3://dlme-metadata-development \
...
 ```
 Note that actual S3 credentials are available from `shared_configs`.
+
+### Configuring transforms
+Configuration for transforms is specified in `metadata_mappings.json`. For example:
+
+```
+[
+  {
+    "traject": "mods_config.rb",
```
jcoyne: I think this should allow multiples.  In DLME I belive we had `traject -c config/traject.rb -c lib/traject/mods_config.rb ...`

invoke.sh
```diff
@@ -14,14 +14,14 @@ if [ $SKIP_FETCH_DATA != "true" ]; then
...
-  mv dlme-metadata-master/* /opt/traject/dlme-metadata/
+  mkdir -p /opt/traject/data
+  mv dlme-metadata-master/* /opt/traject/data
   rm master.zip
 fi
 
 OUTPUT_FILEPATH="output/output-$(date +%Y%m%d%H%M%S).ndjson"
 
-traject -Ilib -w Traject::JsonWriter $@ | tee $OUTPUT_FILEPATH
+exe/transform -d $@ | tee $OUTPUT_FILEPATH
```
jcoyne: can we use `--data-dir` here as it's more communicative?

invoke.sh
```diff
@@ -14,14 +14,14 @@ if [ $SKIP_FETCH_DATA != "true" ]; then
   echo "Getting github.com/sul-dlss/dlme-metadata"
   curl -L https://github.com/sul-dlss/dlme-metadata/archive/master.zip > master.zip
   unzip master.zip
-  mkdir -p /opt/traject/dlme-metadata
-  mv dlme-metadata-master/* /opt/traject/dlme-metadata/
+  mkdir -p /opt/traject/data
+  mv dlme-metadata-master/* /opt/traject/data
```
jcoyne: 👏

metadata_mapping.json
jcoyne: Okay, I can see what you're thinking.

lib/transformer.rb
```diff
@@ -0,0 +1,95 @@
...
+      @mapping_config = mapping_config
+      @base_data_dir = base_data_dir
+      @data_dir = data_dir
+    end
+
+    # Performs the mapping, returning a hash of data filepaths to config filenames
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
```
jcoyne: Can you fix up some of these next week?

PR #39, opened 2019-02-15
include an example of how to run traject locally without docker

peetucket:


README.md
```diff
@@ -85,3 +85,18 @@ docker run --rm -e S3_BUCKET=s3://dlme-metadata-development \
...
+without using docker (so that you can make changes to both the macros and the traject
+configs for testing).
+
+To do this, you need to manually set the Ruby path with the -I switch to include the
+`dlme-transform\lib` directory so the Traject config files can find them with their require statements.
+
+For example:
+
+```
+bundle exec traject -I ../dlme-transform/lib -w DebugWriter -c ../dlme-traject/bnf_cealex_config.rb -s agg_provider='Stanford' -s agg_data_provider='Stanford' ../dlme-metadata/bnf/cealex/data/cealex-99.xml
```
jcoyne: Wouldn't you already be in the `dlme-transform` directory?  Why not `-Ilib` ?

peetucket: good call, changed and rebased

README.md
```diff
@@ -85,3 +85,18 @@ docker run --rm -e S3_BUCKET=s3://dlme-metadata-development \
...
+without using docker (so that you can make changes to both the macros and the traject
+configs for testing).
+
+To do this, you need to manually set the Ruby path with the -I switch to include the
+`dlme-transform\lib` directory so the Traject config files can find them with their require statements.
+
+For example:
+
+```
+bundle exec traject -I lib -w DebugWriter -c ../dlme-traject/bnf_cealex_config.rb -s agg_provider='Stanford' -s agg_data_provider='Stanford' ../dlme-metadata/bnf/cealex/data/cealex-99.xml
```
jcoyne: Just an FYI, if you use the `DebugWriter` you're losing an awful lot of functionality provided by the `DlmeJsonWriter`.  You might want to recommend the latter.

README.md
```diff
@@ -85,3 +85,18 @@ docker run --rm -e S3_BUCKET=s3://dlme-metadata-development \
...
 Note that actual S3 credentials are available from `shared_configs`.
+
+
+For development purposes, if you have the `dlme-traject`, the `dlme-transform`, and the
+`dlme-metadata` repos all checked out locally, you can run the traject transforms locally
+without using docker (so that you can make changes to both the macros and the traject
+configs for testing).
+
+To do this, you need to manually set the Ruby path with the -I switch to include the
+`dlme-transform\lib` directory so the Traject config files can find them with their require statements.
```
jcoyne: drop `dlme-transform\`?

README.md
peetucket: I'll add that as another example in the readme, but curious as to what the difference is.

README.md
peetucket: yup, will clarify a bit more

README.md
peetucket: should be `DlmeJsonResourceWriter` ?

jcoyne: Yes, that's correct.  First of all the `DlmeJsonResourceWriter` writes to JSON, whereas the debug writer just writes a `Hash#to_s`.  Next the `DlmeJsonResourceWriter` casts fields that should be singular (like `id`) from lists. Finally, it runs the DLME IR schema validator on each output to ensure the config is writing a compliant message.

peetucket: Added this  info to readme.

jcoyne:
I'm not entirely sure why we'd want to bother with an alternative way of doing things?  Why not just change the macros (if needed), then `docker build` and then 
```
docker run --rm -e SKIP_FETCH_CONFIG=true \
                -e SKIP_FETCH_DATA=true \
                -v $(pwd)/../dlme-traject:/opt/traject/config \
                -v $(pwd)/../dlme-metadata:/opt/traject/dlme-metadata \
                -v $(pwd)/output:/opt/traject/output \
                suldlss/dlme-transform:latest \
                -c config/mods_config.rb \
                -s agg_provider=Stanford \
                -s agg_data_provider=Stanford \
                dlme-metadata/stanford/maps/data/kj751hs0595.mods
```

My thought is that having two ways of doing this introduces too many things that can go wrong.  I'd like to know more about why you would not use docker.

justinlittman:
I agree with @jcoyne. I think we only want to encourage running transforms using the docker container. Otherwise, it is likely to bite us when some transform works differently in the docker container / in production.

If there is some shortcoming in the docker container (e.g., missing knobs), we can work to add them.

peetucket:
> I agree with @jcoyne. I think we only want to encourage running transforms using the docker container. Otherwise, it is likely to bite us when some transform works differently in the docker container / in production.
> 
> If there is some shortcoming in the docker container (e.g., missing knobs), we can work to add them.

If you are editing and debugging both macros and traject configs together (which I had to do a couple times to debug some issues with missing require statements), having to constantly rebuild the docker container locally simply to run a new test transform started to get annoying.  Also, is it not true that any changes to the macros in `dlme-transform` need to be committed and merged to the master branch in order to rebuild the container?  If so, this is another reason (may want to just test local changes to macros without making remote commits).  Or am I missing something?

jcoyne:
>  Also, is it not true that any changes to the macros in `dlme-transform` need to be committed and merged to the master branch in order to rebuild the container? If so, this is another reason (may want to just test local changes to macros without making remote commits). Or am I missing something?

no, docker is completely unaware of git.

peetucket:
> no, docker is completely unaware of git.

the docker build command in the readme (below) references git, but locally?  doesn't this mean you still need to commit any changes to the macros locally (to master?) to rebuild the docker container? 

 ```
docker build --build-arg VCS_REF=`git rev-parse --short HEAD` \
             --build-arg VCS_URL=`git config --get remote.origin.url` \
             --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
             . -t suldlss/dlme-transform:latest
```

**Comment by jcoyne** [comment]
Those are just labels that we're taking from git and passing to docker to make labels.  They are basically useless unless you are going to publish the image.  You can put whatever. you want there:
```
docker build --build-arg VCS_REF=foo \
            --build-arg VCS_URL=bar \
            --build-arg BUILD_DATE=baz \
            . -t suldlss/dlme-transform:latest
```

justinlittman:
We can make it so that local traject-transform code can be shared into the container (like metadata and configs) so no need to rebuild container to test changes.

PR #42, opened 2019-02-15
Added rubocop.

### LINKED ISSUE #40: Add Rubocop

justinlittman:
closes #40


lib/macros/dlme.rb
```diff
@@ -32,9 +32,7 @@ def first(*macros)
     # only accumulate values if a condition is met
     def conditional(condition, lambda)
```
jcoyne: Ugh. I don't like that we have a parameter that is also a keyword.

PR #44, opened 2019-02-15
add writer class information to readme

jacobthill:


jacobthill:
Ok, I will change it. The other writers do work though, at least DebugWriter does.

jcoyne:
@jacobthill I wouldn't recommend the DebugWriter, because it doesn't do the casting to the final JSON structure or run the output through the IR validator.

PR #50, opened 2019-02-19
align types with application profile

jacobthill:
This can start our conversation about aligning the types translation map with the [application profile](https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md). The name change will break config files that call it but we need to update those anyway so we can hold off on merging until https://github.com/sul-dlss/dlme-metadata/issues/29 is complete.

### DIFF
--- a/lib/translation_maps/edm_types.yaml
+++ b/lib/translation_maps/edm_types.yaml
@@ -3,20 +3,22 @@
 3d: 3d
 artifact: image
 audio: sound
-cartographic: cartographic
-collection: collection
-dataset: dataset
-digital data: dataset
+# cartographic: cartographic
+cartographic: image
+# collection: collection
+# dataset: dataset
+# digital data: dataset
 image: image
-interactive resource: interactive resource
+# interactive resource: interactive resource
 manuscript: text
-map: cartographic
-mixed material: text
+# map: cartographic
+map: image
+# mixed material: text
 moving image: video
-multimedia: [sound, video]
+# multimedia: [sound, video]
 notated music: text
-software: software
-software, multimedia: software
+# software: software
+# software, multimedia: software
 sound: sound
 sound recording: sound
 sound recording-musical: sound
@@ -26,19 +28,21 @@ stillimage: image
 tactile: image
 text: text
 three dimensional object: image
-vector digital data: dataset
+# vector digital data: dataset
 video: video
 # MARC types
 a: text
 c: text
 d: text
-e: cartographic
-f: cartographic
+# e: cartographic
+# f: cartographic
+e: image
+f: image
 g: image
 i: sound
 j: sound
 k: image
-m: software
+# m: software
 p: text
 r: image
 t: text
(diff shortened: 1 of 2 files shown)


jcoyne:
This change seems to break this test: https://github.com/sul-dlss/dlme-transform/blob/c64204e3a748d999d42dfb12a8e87bff54a9f42b/spec/lib/traject/macros/extraction_spec.rb#L45-L51

jacobthill:
Yeah, I'm not sure why but the config files Wayne wrote and the
translation_map do not align with the application_profile. For some unknown
reason he deviated from it. We need to bring this back in accord. I have an
issue open to review all of his config files. We have three type entities
which is possibly why he was confused. I changed the translation_map
'types' to 'edm_types' to avoid any future confusion. I think this is what
is causing the error. We need to update the name everywhere else. But later
when we start using the other cho_has_type in a different translation map
we won't get confused and apply the wrong translation map. I think we need
to hold off on merging until all of these changes are done.

jcoyne:
@jacobthill I'd try changing `spec/lib/traject/macros/extraction_spec.rb` line 46 from `types` to `edm_types`

anarchivist:
My inclination is to keep but change the existing mappings rather than commenting them out. I think a lot of the cartographic stuff will impact the geospatial data that's included there from Stanford and others. While that stuff may be removed eventually, I think we should probably expect that the mappings will still impact what we're doing if we have similar types of materials.

jacobthill:
@anarchivist I'm not sure what you mean. What would collection, dataset and digital data map to, for example?

jacobthill:
@anarchivist Are you talking about the MARC types?

anarchivist:
> Are you talking about the MARC types?

Yes, I am - specifically, in terms of what was deleted/commented out in your changes in this PR. 

> What would collection, dataset and digital data map to, for example?

We could add "collection" and "dataset" as new types to the application profile.

jacobthill:
Ok, i'll add collection and dataset. What about software, interactive resource and cartographic. Do we add those and update the application profile? That is basically how it was before except the application profile didn't reflect what was in the translation_map. If we do that, I have questions about what goes in 'cartographic' and '3d', etc. Europeana has 2d pictures of statues categorized as 3d. I think we should avoid this if possible.

anarchivist:
I'd not include `cartographic`. For what it's worth, the list that i think DPLA uses is based on DCMI type vocabulary values: http://dublincore.org/documents/dcmi-type-vocabulary/#section-7-dcmi-type-vocabulary

Does that sound like a better set of terms to you? I'd be inclined to not include `software`

lib/translation_maps/edm_types.yaml
```diff
@@ -26,19 +28,21 @@ stillimage: image
 tactile: image
 text: text
 three dimensional object: image
-vector digital data: dataset
+# vector digital data: dataset
```
justinlittman: Is there a reason to leave the commented out values instead of removing them?

PR #52, opened 2019-02-19
Adds command_line.filename to index settings.

### LINKED ISSUE #51: Provide command_line.filename to indexer
It's expected by https://github.com/sul-dlss/dlme-transform/blob/master/lib/macros/dlme.rb#L65

justinlittman:
closes #51

### DIFF
--- a/lib/transformer.rb
+++ b/lib/transformer.rb
@@ -89,6 +89,7 @@ module Dlme
       @transformer ||= Traject::Indexer.new.tap do |indexer|
         config_filepaths.each { |config_filepath| indexer.load_config_file(config_filepath) }
         indexer.settings do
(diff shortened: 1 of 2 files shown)


lib/transformer.rb
```diff
@@ -89,6 +89,7 @@ def transformer
       @transformer ||= Traject::Indexer.new.tap do |indexer|
         config_filepaths.each { |config_filepath| indexer.load_config_file(config_filepath) }
         indexer.settings do
+          provide 'command_line.filename', this.input_filepath
```
jcoyne: Is this going to work with wildcards?

justinlittman: I believe so because a new transformer is created for each file.

PR #57, opened 2019-02-20
Adds optional SNS notification containing results summary.

### LINKED ISSUE #33: Notify the webapp when done
Ref https://github.com/sul-dlss/dlme/issues/410

We need to provide status, any errors and the s3 url.

justinlittman:
closes #33

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -3,6 +3,8 @@
 require 'json'
 require 'thor'
 require 'transformer'
+require 'record_counter'
+require 'date'
 
 module Dlme
   module CLI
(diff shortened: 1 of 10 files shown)


justinlittman:
Still needs tests.

justinlittman:
Yippee!
screen shot 2019-02-20 at 3 34 33 pm

lib/cli.rb
```diff
@@ -3,6 +3,8 @@
 require 'json'
 require 'thor'
 require 'transformer'
+require 'record_counter'
+require 'Date'
```
jcoyne: this should be `date`.  See the example: https://ruby-doc.org/stdlib-2.3.3/libdoc/date/rdoc/DateTime.html

lib/transformer.rb
```diff
@@ -74,12 +74,12 @@ def initialize(input_filepath:, config_filepaths:, settings: {})
...
     end
 
-    # Transform a stream into a new representation, using Traject
+    # Transform a stream into a new representation, using Traject, and returns the number of transformed records.
     def transform
       transformer.process(File.open(input_filepath, 'r'))
     rescue RuntimeError => e
       warn "[ERROR] #{e.message}"
-      exit(1)
+      raise
```
jcoyne: Do we even need this rescue block? I think we can just handle this in `cli.rb`

PR #63, opened 2019-02-22
Adds data path and duration to notification message.

### LINKED ISSUE #62: Add duration and data_dir to transform notification
Because they would be helpful.

justinlittman:
closes #62

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -44,6 +44,7 @@ module Dlme
       # rubocop:disable Metrics/MethodLength
       # rubocop:disable Metrics/AbcSize
       def transform
+        @start = Time.now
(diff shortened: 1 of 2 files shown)


lib/cli.rb
```diff
@@ -72,19 +73,25 @@ def mapping
         @mapping ||= JSON.parse(File.read(options.fetch(:mapping_file)))
       end
 
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/AbcSize
       def write_summary(error: nil)
         return unless options.key?(:summary_filepath)
 
         result = {
```
jcoyne: Can you extract this into a `build_summary` method?

justinlittman: Fixed.

PR #64, opened 2019-02-22
Mappings

jacobthill:


lib/macros/dlme.rb
```diff
@@ -4,9 +4,10 @@ module Macros
   # DLME helpers for traject mappings
   module DLME
     LOC_NS = {
-      srw: 'http://www.loc.gov/zing/srw/',
+      dc: 'http://purl.org/dc/elements/1.1/',
       oai_dc: 'http://www.openarchives.org/OAI/2.0/oai_dc/',
-      dc: 'http://purl.org/dc/elements/1.1/'
+      srw: 'http://www.loc.gov/zing/srw/',
+      xmlns: 'http://www.openarchives.org/OAI/2.0/'
```
jcoyne: Would this apply to every XML file?  Is that something we want?

jacobthill: I'm not sure about that. It seems like this shouldn't go in the config files since it will be used by many of them and I don't want to repeat it. This is where @peetucket put the other namespaces so I added xmlns here. Feel free to put it somewhere else if you think its best.

jcoyne: xmlns is the default namespace.  I'm concerned that `http://www.openarchives.org/OAI/2.0/` is the default namespace for OAI-PMH, but not all of the types of XML content that we have to process are OAI-PMH data.  Take a look at `harvard/maps/data/HARVARD.SDE2.G7400_1690_A5_COPY1.fgdc.xml` for example.

peetucket: yeah, to be safe, we should move this to a provider specific namespace variable in the config file in the `dlme-traject` directory (could be a common file if multiple configs are necessary for a specific provider, or just the single config file for that provider).  If the provider specific namespace declaration includes everything from the common `LOC_NS` and just adds the `xlmns` declaration, we can avoid repetition by doing something like this: ```NEW_NS = {      xmlns: 'http://www.openarchives.org/OAI/2.0/' }.merge(LOC_NS).freeze```

PR #66, opened 2019-02-22
Adds patched debug writer. Adds debug writer as commandline option.

### LINKED ISSUE #65: Write version of DebugWriter
We need to write our own version of DebugWriter. There is an error in LineWriter https://github.com/traject/traject/pull/202 that is causing problems when trying to run on a directory of files. I use DebugWriter to check that the transformations I'm making are doing what they are supposed to. I understand that DebugWriter will not check the IR. I will use JsonWriter for that once I'm happy with the transformations.

justinlittman:
closes #65

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -40,6 +40,12 @@ module Dlme
              desc: 'Filepath containing summary of transformation.',
              aliases: '-s'
 
+      option :debug_writer,
+             type: :boolean,
+             banner: 'DEBUG_WRITER',
+             desc: 'Use the debug writer.',
+             aliases: '-w'
+
       desc 'transform', 'Perform a transform'
       # rubocop:disable Metrics/MethodLength
       # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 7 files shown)


lib/cli.rb
```diff
@@ -50,10 +56,12 @@ def transform
           data_dir: options.fetch(:data_dir)
         ).map
         data_filepath_mapping.each do |data_filepath, config|
+          puts(options[:debug_writer])
```
jcoyne: Do we need this here?

lib/dlme_debug_writer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'traject'
+
+# Temporary fix until https://github.com/traject/traject/pull/202 is merged.
+class DlmeDebugWriter < Traject::DebugWriter
+  def close
```
jcoyne: Could we pull this into a module to be mixed into DlmeJsonResourceWriter & DlmeDebugWriter?

lib/cli.rb
justinlittman: Nope. Removed.

PR #77, opened 2019-02-27
Add macro for namespaces from Bibliothèque nationale de France

### LINKED ISSUE #72: Extract a BNF macro
This should at a minimum contain the namespaces that are currently in `Macros::DLME::LOC_NS`

jcoyne:
Fixes #72


justinlittman:
Does this require a change in dlme-traject as well?

jcoyne:
Yes, the companion is https://github.com/sul-dlss/dlme-traject/pull/14

PR #78, opened 2019-02-27
Fix MetThumbnailFetcher

jcoyne:
Fixes  sul-dlss/dlme-traject#10

### DIFF
--- a/lib/dlme_utils.rb
+++ b/lib/dlme_utils.rb
@@ -1,15 +1,16 @@
 # frozen_string_literal: true
 
 require 'active_support/benchmarkable'
+require 'faraday'
 
 module DLME
   # Misc. utilities for working with DLME data
   module Utils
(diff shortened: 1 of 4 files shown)


lib/dlme_utils.rb
```diff
@@ -1,25 +1,29 @@
...
 module DLME
   # Misc. utilities for working with DLME data
   module Utils
     extend ActiveSupport::Benchmarkable
 
-    def self.caching_faraday_client
+    def self.client
       Faraday.new do |builder|
-        builder.use :http_cache, store: Rails.cache
+        # builder.use :http_cache, store: Rails.cache
```
justinlittman: Cruft?

jcoyne: I left this in here to show how we could do caching in the future if we need it.

PR #83, opened 2019-02-27
Provide unmapped language value in addition to NOT_FOUND.

### LINKED ISSUE #45: Don't pass through raw translation values if none match.
Please provide:
What steps should I take to duplicate this problem?
What is the incorrect behavior you are seeing?
What is the correct behavior you want to see?

justinlittman:
closes #45


justinlittman:
Test added @jcoyne

PR #86, opened 2019-02-28
Removes metadata_mapping.json

### LINKED ISSUE #85: Move metadata_mappings.json to dlme-traject
This avoids the need of a change to this repo when adding a collection.

justinlittman:
closes #85


spec/lib/config_finder_spec.rb
```diff
@@ -27,7 +27,7 @@
     ]
   end
 
-  let(:mapping_file) { 'metadata_mapping.json' }
+  let(:mapping_file) { 'spec/lib/metadata_mapping.json' }
```
jcoyne: Can this be `spec/fixtures`?  `spec/lib` would be the place to put tests for code that's in `./lib`

justinlittman: Done.

PR #95, opened 2019-02-28
Encode the prefix into the srw macro

jcoyne:


lib/macros/srw.rb
```diff
@@ -9,11 +9,14 @@ module SRW
       srw: 'http://www.loc.gov/zing/srw/'
     }.freeze
 
+    PREFIX = 'srw:record/srw:recordData/oai_dc:dc/'.freeze
```
justinlittman: Is freeze necessary? Aren't strings frozen by default?

lib/macros/srw.rb
```diff
@@ -9,11 +9,14 @@ module SRW
...
+    private_constant :PREFIX
+
     # Extracts values for the given xpath
     # @param [String] xpath the xpath query expression
     # @return [Proc] a proc that traject can call for each record
     def extract_srw(xpath, options = {})
-      extract_xml(xpath, NS, options)
+      extract_xml("#{PREFIX}#{xpath}", NS, options)
```
justinlittman: That this is prefixed should be noted in doc comments.

lib/macros/srw.rb
jcoyne: I thought so, but the version of rubocop wanted it this way. *shrug*

jcoyne: See #97

lib/macros/srw.rb
jcoyne: 👍

PR #96, opened 2019-02-28
Add an OAI macro

jcoyne:


lib/macros/oai.rb
```diff
@@ -0,0 +1,25 @@
...
+      oai: 'http://www.openarchives.org/OAI/2.0/',
+      dc: 'http://purl.org/dc/elements/1.1/',
+      oai_dc: 'http://www.openarchives.org/OAI/2.0/oai_dc/'
+    }.freeze
+    private_constant :NS
+
+    PREFIX = '/oai:record/oai:metadata/oai_dc:dc/'.freeze
+    private_constant :PREFIX
+
+    # Extracts values for the given xpath
```
justinlittman: Docs should note that prefix applied.

lib/macros/oai.rb
```diff
@@ -0,0 +1,25 @@
...
+  # Macros for extracting OAI values from Nokogiri documents
+  module OAI
+    NS = {
+      oai: 'http://www.openarchives.org/OAI/2.0/',
+      dc: 'http://purl.org/dc/elements/1.1/',
+      oai_dc: 'http://www.openarchives.org/OAI/2.0/oai_dc/'
+    }.freeze
+    private_constant :NS
+
+    PREFIX = '/oai:record/oai:metadata/oai_dc:dc/'.freeze
```
justinlittman: Aren't strings already frozen?

PR #98, opened 2019-03-01
Mappings

jacobthill:


lib/translation_maps/types.yaml
```diff
@@ -43,18 +48,21 @@ Baskı: text
...
 Gazete küpürü: text
 Gravür: image
 Hat: text
+Karikatür: image
 Kartpostal: image
 Kitap bölümü: text
 Kumaş: image
 Köşe yazısı: text
-Makale: text
+Makale:
```
anarchivist: This is unset in the latest revision - was this intentional?

lib/translation_maps/types.yaml
```diff
@@ -1,19 +1,23 @@
...
 image: image
 interactive resource: interactive resource
 manuscript: text
-map: cartographic
+map: image
 mixed material: text
 moving image: video
 multimedia: [sound, video]
 notated music: text
+software: software
```
anarchivist: I'm not super picky about this but I wonder if we should just use `interactive resource` here.

jacobthill:
I'm still getting confused with multiple repos and branches going at once. software should be set to interactive resource and Makale should be set to text. I will fix those.

PR #100, opened 2019-03-01
Rescue all errors, not just RuntimeError.

### LINKED ISSUE #99: Transform swallows errors
Steps to reproduce:
```
docker run --rm -e SKIP_FETCH_CONFIG=true                 -e SKIP_FETCH_DATA=true                 -v $(pwd)/.:/opt/traject                 -v $(pwd)/../dlme-traject:/opt/traject/config                 -v $(pwd)/../dlme-metadata:/opt/traject/data                 -v $(pwd)/output:/opt/traject/output                 suldlss/dlme-transform:latest
```

At present, this raises a TypeError.

Expected result:
Error is caught and noted in summary document.

Actual result:
Summary document is not written. User has no indication that failed.

The transformer should caught and log all errors, not just RuntimeErrors raised by validation.

justinlittman:
closes #99

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -52,11 +52,13 @@ module Dlme
         @start = Time.now
         transform_all
         write_summary
-      rescue RuntimeError => e
+      # rubocop:disable Style/RescueStandardError
+      rescue => e
         warn "[ERROR] #{e.message}"
         write_summary(error: e.message)
         exit(1)
       end
+      # rubocop:enable Style/RescueStandardError
 
       default_task :transform
 


lib/cli.rb
```diff
@@ -52,11 +52,13 @@ def transform
         @start = Time.now
         transform_all
         write_summary
-      rescue RuntimeError => e
+      # rubocop:disable Style/RescueStandardError
```
jcoyne: I like #101 better because it doens't have any rubocop annotations in it.

PR #116, opened 2019-03-12
Updated Met macros for JSON instead of CSV.

justinlittman:
references https://github.com/sul-dlss/dlme-traject/issues/27


jcoyne:
We no longer need `met_thumbnail`?

PR #120, opened 2019-03-12
Use extract_xpath (traject) instead of extract_xml (traject_plus) 

jcoyne:


lib/macros/oai.rb
```diff
@@ -13,13 +13,17 @@ module OAI
     PREFIX = '/oai:record/oai:metadata/oai_dc:dc/'
     private_constant :PREFIX
 
+    def self.extended(mod)
```
justinlittman: Try to understand self.extended. Why not just extend OAI module with Traject::Macros::NokogiriMacros?

jcoyne: AFAIK:  Modules can only extend classes, not modules

PR #124, opened 2019-03-19
add values found in auc to translation maps

jacobthill:
Let me know if you have other thoughts on how to deal with errors in the metadata. I found `eg` and `englishglish` for language and mapped them to nothing so I could get past the not_found test. `englishglish` clearly means English but that was already added so removing `englishglish` won't take 'English' out in this case

### DIFF
--- a/lib/translation_maps/languages.yaml
+++ b/lib/translation_maps/languages.yaml
@@ -205,3 +205,7 @@ slovakça: Slovak
 türkçe: Turkish
 türkçe-osmanlıca: Turkish, Ottoman
 yunanca: Greek
+
+# Errors: map to nothing to pass not_found test
+eg:
+englishglish:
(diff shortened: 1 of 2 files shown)


lib/translation_maps/languages.yaml
```diff
@@ -205,3 +205,7 @@ slovakça: Slovak
 türkçe: Turkish
 türkçe-osmanlıca: Turkish, Ottoman
 yunanca: Greek
+
+# Errors: map to nothing to pass not_found test
```
anarchivist: My inclination is to separate these into their own map. The behavior is different enough here that I'm hoping we can encapsulate it. I'd also consider doing the same with the Turkish language map.

jacobthill: @anarchivist I broke both out into separate files

anarchivist: These lines still appear in this file

PR #131, opened 2019-04-02
write mods macro

jacobthill:
The mods.rb file that was there was not used by anything as far as I can tell. I think it was an old macro that Wayne wrote. I have a WIP config that uses this macro here: https://github.com/sul-dlss/dlme-traject/pull/new/qnl-config

I think this is right but let me know if I missed anything.


jacobthill:
Actually I just noticed a mods_config.rb file that uses this. Let me rethink this

PR #132, opened 2019-04-03
Fixes adjust_cardinality to handle agg_has_view arrays. Fixes marc da…

### LINKED ISSUE #128: Fix cho_date and agg_has_view "must be string" errors in harvard and penn data
Try to run the current configs on 'penn' or 'harvard' for specific examples

justinlittman:
…tes to make strings.

closes #128

### DIFF
--- a/lib/adjust_cardinality.rb
+++ b/lib/adjust_cardinality.rb
@@ -33,13 +33,23 @@ class AdjustCardinality
   end
 
   def flatten_web_resources(attributes)
-    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview]
+    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview agg_has_view]
     attributes.except(*flatten).tap do |output|
       flatten.each do |field|
-        next unless attributes.key?(field)
+        value = attributes[field]
+        next unless value
 
-        output[field] = process_web_resource(attributes.fetch(field))
+        output[field] = flatten_web_resource(value)
       end
+    end
+  end
+
+  def flatten_web_resource(web_resource)
+    # For handling agg_has_view, which is an array.
+    if web_resource.is_a?(Array)
+      web_resource.map { |wr| process_web_resource(wr) }
+    else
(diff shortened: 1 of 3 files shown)


lib/adjust_cardinality.rb
```diff
@@ -32,16 +32,23 @@ def flatten_top_level(attributes)
...
   def flatten_web_resources(attributes)
-    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview]
+    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview agg_has_view]
     attributes.except(*flatten).tap do |output|
       flatten.each do |field|
         next unless attributes.key?(field)
 
-        output[field] = process_web_resource(attributes.fetch(field))
+        # If agg_has_view, this is an array
+        output[field] = if attributes.fetch(field).is_a?(Array)
```
jcoyne: can you extract this to a method?

lib/adjust_cardinality.rb
```diff
@@ -32,16 +32,23 @@ def flatten_top_level(attributes)
     end
   end
 
+  # rubocop:disable Metrics/MethodLength
   def flatten_web_resources(attributes)
-    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview]
+    flatten = %w[agg_is_shown_at agg_is_shown_by agg_preview agg_has_view]
     attributes.except(*flatten).tap do |output|
       flatten.each do |field|
         next unless attributes.key?(field)
```
jcoyne: If we do
```ruby
value = attributes[field]
next unless value
```
then you don't need to call `attributes.fetch(field)` below.

lib/adjust_cardinality.rb
justinlittman: I thought about that but couldn't come up with a name for a method that was less confusing than just if/then code. Suggestions?

jcoyne: I'd call it `flatten_web_resource`  as this is for one item in the list.

PR #136, opened 2019-04-04
Qnl macro

jacobthill:


lib/macros/qnl.rb
```diff
@@ -0,0 +1,36 @@
...
+    def self.extended(mod)
+      mod.extend Traject::Macros::NokogiriMacros
+    end
+
+    # Extracts values for the given xpath which is prefixed with oai and mods wrappers
+    # @example
+    #   extract_mods('mods:language') => lambda { ... }
+    # @param [String] xpath the xpath query expression
+    # @return [Proc] a proc that traject can call for each record
+    def extract_mods(xpath)
```
jcoyne: There is already an `extract_mods` macro provided by traject plus.  I think you may want to name this something else to avoid confusion.

PR #143, opened 2019-04-09
Refactor CONTENTdm macros to get procs (fixes #141)

### LINKED ISSUE #141: Refactor content_dm macro
The `normalize` function takes a string but traject typically needs the second argument to return a proc/lambda.  Calling normalize like so: ```to_field 'agg_preview', thumbnail(extract_oai('dc:identifier[last()]'))``` results in ```NoMethodError:undefined method `include?' for #<Proc:0x000056073c34bdd8>```. Need a way to pass the extracted url to `normalize`

anarchivist:
Ref sul-dlss/dlme-traject#59

### DIFF
--- a/lib/macros/content_dm.rb
+++ b/lib/macros/content_dm.rb
@@ -1,15 +1,43 @@
 # frozen_string_literal: true
 
+require 'traject_plus'
+
 module Macros
   # Macros for working with CONTENTdm data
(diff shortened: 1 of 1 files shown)


jacobthill:
@jcoyne is this ok to merge? I need it to transform the Sakip Sabanci in stage

PR #149, opened 2019-04-17
Use the logger instead of puts

aaron-collier:
Instead put just doing a `puts` on the error, use the logger.

### DIFF
--- a/lib/dlme_utils.rb
+++ b/lib/dlme_utils.rb
@@ -27,7 +27,7 @@ module DLME
     end
 
     def self.logger
-      @logger ||= Logger.new(STDOUT)
+      @logger ||= Logger.new(STDERR)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/dlme_utils.rb
```diff
@@ -27,7 +27,7 @@ def self.fetch_json(uri)
     end
 
     def self.logger
-      @logger ||= Logger.new(STDOUT)
+      @logger ||= Logger.new(STDERR)
```
jcoyne: why do we need to change this if we're not using the logger in lib/macros/iiif.rb?

aaron-collier: The benchmark debug gets stuck into the output file as something there (traject?) is capturing STDOUT. This is just a quick fix to unblock @jacobthill and I'll walk through updating this today.

PR #170, opened 2019-07-30
add macro for Michigan data

jacobthill:


jacobthill:
I'm not sure how to write a test for this. Do you have an example? Also, the value is tested in use (e.g. when this macro is used on the cho_language field, I also call the not_found test @justinlittman wrote to ensure that the result is a valid language and is found in one of the language translation maps).

jcoyne:
Call the method that is returned by `last_only` and make sure it mutates the arguments passed to it as expected.

jacobthill:
If I understand you correctly, I think that is done [here](https://github.com/sul-dlss/dlme-traject/blob/master/michigan_config.rb)

PR #206, opened 2019-09-25
fix dlme_json_schema I hope

ndushay:
## Why was this change made?

When performing `docker run` on data formerly working without a hitch, there were errors like

```
The errors are: [#<Dry::Schema::Message text="is missing" path=[:cho_title] predicate=:key? input={}>, #<Dry::Schema::Message text="is missing" path=[:id] predicate=:key? input={}>, #<Dry::Schema::Message text="is missing" path=[:agg_data_provider] predicate=:key? input={}>, #<Dry::Schema::Message text="is missing" path=[:agg_provider] predicate=:key? input={}>]}
``` 

This fix avoids the errors.  Whether or not the desired validation is happening may need the help of @justinlittman eventually, but should not hold up this PR.

n/a

### DIFF
--- a/lib/dlme_json_schema.rb
+++ b/lib/dlme_json_schema.rb
@@ -5,23 +5,22 @@ require 'dry-schema'
 # rubocop:disable Metrics/BlockLength
 
 # See https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md#svcsservice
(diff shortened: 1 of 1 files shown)


ndushay:
Merging to unblock Jacob;  happy to address anything with follow-on PRs.

PR #217, opened 2019-09-26
dlme_json_schema: some fixes from upgrade to dry-schema gem

ndushay:
#### stanford data was getting these dry-schema errors:

```
[ERROR] Transform produced invalid data.

The errors are: [
#<Dry::Schema::Message text="is missing" path=[:agg_is_shown_by, :wr_has_service, 0, :service_id] predicate=:key? 
input={"service_conforms_to"=>["http://iiif.io/api/image/"], "service_id"=>"https://stacks.stanford.edu/image/iiif/cm881bj1960%2FW705_000001_300", "service_implements"=>"http://iiif.io/api/image/2/level2.json"}>, 

#<Dry::Schema::Message text="is missing" path=[:agg_is_shown_by, :wr_has_service, 0, :service_conforms_to] predicate=:key? 
input={"service_conforms_to"=>["http://iiif.io/api/image/"], "service_id"=>"https://stacks.stanford.edu/image/iiif/cm881bj1960%2FW705_000001_300", "service_implements"=>"http://iiif.io/api/image/2/level2.json"}>, 

#<Dry::Schema::Message text="is missing" path=[:agg_preview, :wr_has_service, 0, :service_id] predicate=:key? 
input={"service_conforms_to"=>["http://iiif.io/api/image/"], "service_id"=>"https://stacks.stanford.edu/image/iiif/cm881bj1960%2FW705_000001_300", "service_implements"=>"http://iiif.io/api/image/2/level2.json"}>, 

#<Dry::Schema::Message text="is missing" path=[:agg_preview, :wr_has_service, 0, :service_conforms_to] predicate=:key? 
input={"service_conforms_to"=>["http://iiif.io/api/image/"], "service_id"=>"https://stacks.stanford.edu/image/iiif/cm881bj1960%2FW705_000001_300", "service_implements"=>"http://iiif.io/api/image/2/level2.json"}>]}
```

#### This PR fixes the above
and also uses a more compact way to express "element x should be an array of string values (see "each" explained here:  https://dry-rb.org/gems/dry-schema/basics/macros/ )

Note that I'm not sure if "excluded_from" is correct in our schema now, but ... maybe???   Let's watch for those errors on xform.

n/a

### DIFF
--- a/lib/dlme_json_schema.rb
+++ b/lib/dlme_json_schema.rb
@@ -7,57 +7,57 @@ require 'dry-schema'
 # See https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md#svcsservice
 ServicesSchema = Dry::Schema.JSON do
   required(:service_id).filled(:string)
-  required(:service_conforms_to) { array? { each(:str?) } }
+  required(:service_conforms_to).array(:str?)
   optional(:service_implements).filled(:string)
 end
 
 # See https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md#edmwebresource
 EDMWebResourceSchema = Dry::Schema.JSON do
   required(:wr_id).filled(:string)
-  optional(:wr_format) { array? { each(:str?) } }
-  optional(:wr_has_service).each do
+  optional(:wr_format).array(:str?)
+  optional(:wr_has_service).value(:array).each do
     schema(ServicesSchema)
   end
 end
 
 DlmeJsonSchema = Dry::Schema.JSON do
   # See https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md#edmprovidedcho
-  optional(:cho_alternative) { array? { each(:str?) } }
-  optional(:cho_contributor) { array? { each(:str?) } }
-  optional(:cho_coverage) { array? { each(:str?) } }
-  optional(:cho_creator) { array? { each(:str?) } }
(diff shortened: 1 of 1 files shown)


lib/dlme_json_schema.rb
```diff
@@ -7,57 +7,57 @@
...
-  optional(:cho_dc_rights) { array? { each(:str?) } }
-  optional(:cho_description) { array? { each(:str?) } }
+  optional(:cho_alternative).array(:str?)
+  optional(:cho_contributor).array(:str?)
+  optional(:cho_coverage).array(:str?)
+  optional(:cho_creator).array(:str?)
+  optional(:cho_date).array(:str?)
+  optional(:cho_dc_rights).array(:str?)
+  optional(:cho_description).array(:str?)
   optional(:cho_edm_type) { array? { each(:str?, excluded_from?: ['NOT FOUND']) } }
```
ndushay: the 2 lines like this are the ones I'm not clear on;  I tried to read the documentation, but couldn't puzzle it out for our example.

I also searched dlme-metadata for "NOT FOUND" and didn't get any hits, so maybe it doesn't matter.

mjgiarlo: I think it's saying that every member of the `cho_edm_type` array must be a string and must not equal "NOT FOUND"

ndushay: I was more concerned as to whether the syntax was correct, given all the other bumps were with syntax ...

PR #219, opened 2019-09-27
closes #215

### LINKED ISSUE #215: Rename penn_thumbnail to avoid future confusion
This method only works on the openn subdirectory, not the others. To avoid misusing it in the future it should be renamed to `openn_thumbnail`.

jacobthill:
## Why was this change made?
Need to map an earliest date and latest date from separate incoming fields to a single `cho_date_norm` range. 

Not applicable

### DIFF
--- a/lib/macros/normalize.rb
+++ b/lib/macros/normalize.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for normalizing incoming metadata
+  module Normalize
+    # Extracts earliest & latest dates and merges into singe date range value
+    # @example
+    #   penn_museum_date_range => lambda { ... }
+    # @return [Proc] a proc that traject can call for each record
+    def penn_museum_date_range
+      lambda do |record, accumulator, _context|
+        accumulator << [record['date_made_early'], record['date_made_late']].select(&:present?).join('/').presence
+      end
+    end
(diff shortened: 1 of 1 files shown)


lib/macros/normalize.rb
```diff
@@ -0,0 +1,16 @@
...
+module Macros
+  # Macros for normalizing incoming metadata
+  module Normalize
+    # Extracts earliest & latest dates and merges into singe date range value
+    # @example
+    #   penn_museum_date_range => lambda { ... }
+    # @return [Proc] a proc that traject can call for each record
+    def penn_museum_date_range
+      lambda do |record, accumulator, _context|
+        accumulator << [record['date_made_early'], record['date_made_late']].select(&:present?).join('/').presence
```
ndushay: so is the eventual idea that the stuff I'm working on will go in here -- that after parsing out the earliest date and the latest date from the record (which may be from stuff I'm working on), we then turn it into a range with the method I wrote in branch year_array?   

Is that how this will eventually go together?

And some collections have the earliest year and the latest year in ONE field coming in, and some already have them separated?

jacobthill: yes, and if they are separated we will merge them so you will always have a single field to work with, a field that could contain single year dates, ranges, or centuries, etc.

PR #220, opened 2019-09-27
Add cho_date_norm and cho_date_hijri to schema

aaron-collier:
## Why was this change made?

The IR requires a field for the conversion of dates into hijri dates. Additional we need an array to capture the integer representation of all individual years within a range.

N/A

### DIFF
--- a/lib/dlme_json_schema.rb
+++ b/lib/dlme_json_schema.rb
@@ -28,7 +28,7 @@ DlmeJsonSchema = Dry::Schema.JSON do
   optional(:cho_creator).array(:str?)
   optional(:cho_date).array(:str?)
   optional(:cho_date_hijri).array(:str?)
-  optional(:cho_date_norm).array(:integer) # this is an array of integers specificaly to capture the int representaiton of all years in a range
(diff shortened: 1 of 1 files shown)


lib/dlme_json_schema.rb
```diff
@@ -27,6 +27,8 @@
   optional(:cho_coverage).array(:str?)
   optional(:cho_creator).array(:str?)
   optional(:cho_date).array(:str?)
+  optional(:cho_date_hijri).array(:str?)
+  optional(:cho_date_norm).array(:integer) # this is an array of integers to capture the years in a range
```
ndushay: would this be better named cho_date_range_norm

ndushay: of course, it would need to match sul-dlss/dlme-traject/pull/128

aaron-collier: Yeah, probably, will rename.

aaron-collier:
@ndushay Something to verify with @jacobthill but I think we'll need it in solr for faceting. I could be wrong.

PR #222, opened 2019-09-27
[HOLD UNTIL MONDAY SEP 30] validate existence of new fields coming from macro

peetucket:
## Why was this change made?

part of sul-dlss/dlme-traject#111

n/a


peetucket:
@jacobthill : just double-checking that these new fields are required ... and that therefore validation will fail for new transforms if we merge this until the metadata_mappings.json file in dlme-traject is updated

jacobthill:
Lets hold off on merging for now. I need to make some transformations this afternoon and won't have time to add the country values until Monday morning.

PR #231, opened 2019-10-01
add date_parsing macros

ndushay:
## Why was this change made?

DLME date slider requires a Solr field with an Array of integer years for the whole range possible in a record.  This PR contains low hanging fruit for parsing out integer dates (and ranges) from strings using 2 existing gems, which will enable appropriate values for the date slider Solr field for about 8 collections.

These methods may be modified in the future to accommodate more data, and may not be relying on unmaintained gems after that.

Note that yes, it is possible to test our lambdas (and we should!)

n/a

### DIFF
--- a/lib/macros/date_parsing.rb
+++ b/lib/macros/date_parsing.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'timetwister'
+require 'stanford-mods'
+
+# Macros for Traject transformations.
+module Macros
+  # Macros for parsing dates from Strings
+  module DateParsing
+    # get array of year values in range, when string is:
+    # yyyy-yyyy
+    # yyyy - yyyy (can be one or more spaces by hyphen, but not other types of whitespace)
+    # yyyy  (one element in result)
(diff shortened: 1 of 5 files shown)


lib/macros/date_parsing.rb
```diff
@@ -0,0 +1,35 @@
...
+    # yyyy  (one element in result)
+    #  will not work for negative numbers, or fewer than 4 digit years
+    def range_array_from_positive_4digits_hyphen
+      lambda do |_record, accumulator|
+        range_years = []
+        accumulator.each do |val|
+          range_years << Timetwister.parse(val).first[:index_dates]
+        end
+        range_years.flatten!.uniq!
+        accumulator.replace(range_years)
```
mjgiarlo: @ndushay I wonder if this could be shortened to:

```ruby
def range_array_from_positive_4digits_hyphen
    accumulator.map! { |val| parse(val) }.flatten!.uniq!
  end
end

def parse(val)
  Timetwister.parse(val).first[:index_dates]
end
```

:question:

ndushay: good suggestion!

Issue #173, opened 2019-08-08
Modify intermediate representation to support language tags for metadata field values


jacobthill:
Requirements:
- [x] add language of property values to IR (Look at IIIF for multilingual modeling in JSON v2 or v3? https://iiif.io/api/presentation/3.0/#44-language-of-property-values ) 
- [x] write macro to manually assign language values to each field  (e.g. `to_field 'cho_creator', extract_oai('dc:creator'), lang('ara')`
- [x] write test to ensure that the language value passed through the config is a valid BCP47 language value

mjgiarlo:
I believe all the components of this work are now done. See the list of issues and commits above.

PR #240, opened 2019-10-03
Support tagging metadata values with language codes

mjgiarlo:
Connects to #173

Includes:
* Switch from dry-schema to dry-validation for more robust validation features
* Rework DLME schemas as dry-validation contracts
* Store list of acceptable BCP47 language codes in a YAML file to validate string values passed to `lang()` Traject macro
* Validate `cho_title` field as a hash with keys that are acceptable BCP47 codes
* Collapse services schema into EDM web resource schema
* Add Traject `lang()` macro that takes a single BCP47 language code
* Add post-processing macro for converting mapping values to a IIIF-flavored language hash

**NOTE**: For now we are only language-tagging the title field to test/vet things. Once we decide to open this up to other fields, we'll need to update `Contracts::CHO`. 

The whole journey of language-tagged fields goes something like this:

1. First, a macro like `extract_xpath` is used to map data out of the source record and begins its life as a string in an array: `['title1', 'title2']`
1. Then, the `lang()` macro, if used, takes the array of strings and reworks it into an array of hashes, with each hash representing values in a given language: `[{ language: 'en', values: ['title1', 'title2'] }]`
1. Finally, the `convert_to_language_hash(field1, field2, etc.)` macro reworks the above array into a IIIF-flavored hash after all mappings are complete, so from `[{ language: 'en', values: ['title1', 'title2'] }, { language: 'ar-Arab', values: ['arabic title1', 'arabic title2'] }]` to `{ 'en' => ['title1', 'title2'], 'ar-Arab' => ['arabic title1', 'arabic title2'] }`

:point_up: @jacobthill @aaron-collier


jacobthill:
@mjgiarlo I think this needs the Arabic fields that were added in https://github.com/sul-dlss/dlme-transform/commit/412fb42f9ddcd2645943077a8290d6ba2936728b

jacobthill:
Also I think I need to catch up on the decision to only require title as a hash value. The implications are that we can't assign language values to any other field. The normalized fields have Arabic equivalents so this is partly solved for them but for cases where the description, creator, contributor, etc. is in Arabic we have no way of indicating that in the data model. This will place design constraints on @ggeisler. I'm not sure we are ready to commit to this route. We can, of course, allow for the language of other fields to be captured later with further changes. Was that the idea?

jacobthill:
@mjgiarlo addressed all my concerns in person. I think this plan makes sense.

PR #249, opened 2019-10-07
Ans dates

jacobthill:
## Why was this change made?
To normalize the ANS dates for date slider

NA

### DIFF
--- a/lib/macros/date_parsing.rb
+++ b/lib/macros/date_parsing.rb
@@ -45,7 +45,7 @@ module Macros
     def american_numismatic_date_range
       lambda do |record, accumulator|
         if record['Year']
-          dates = record['Year'].split('|') #if record['Year']
+          dates = record['Year'].split('|')
           first_year = dates[0].to_i if dates[0]&.match(/\d+/)
           last_year = dates[1].to_i if dates[1]&.match(/\d+/)
         end
--- a/lib/macros/date_parsing.rb
+++ b/lib/macros/date_parsing.rb
@@ -41,6 +41,18 @@ module Macros
       end
     end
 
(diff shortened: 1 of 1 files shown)


jacobthill:
@aaron-collier I'm working on writing tests. I got pulled into something else this afternoon. Can we find a time to pair on this tomorrow?

lib/macros/date_parsing.rb
```diff
@@ -41,6 +41,18 @@ def penn_museum_date_range
       end
     end
 
+    # Extracts earliest & latest dates from American Numismatic Society record and merges into singe date range value
+    def american_numismatic_date_range
+      lambda do |record, accumulator|
+        if record['Year']
+          dates = record['Year'].split('|')
+          first_year = dates[0].to_i if dates[0]&.match(/\d+/)
```
ndushay: Your regex will not work for negative years, I believe.  I don't think penn museum had any negative years, and I didn't anticipate that.

It would be a good idea to create a constant like

```
YEAR_REGEX = /-?\d+/
```

and to use it consistently (ans and penn and whatever else comes in)

```
first_year = dates[0].to_i if dates[0]&.strip&.match(YEAR_REGEX)
```

ndushay: Also, here you could use the accumulator to get the value, as the data is only coming in from a single field.

```
  accumulator.map! do |val|
        first_year = ...
        last_year = ...
        Macros::DateParsing.year_array(first_year, last_year)
   end
end
```

`.map!` iterates through each original value and will change the field value in place.  This is what we want here.   Again, I couldn't go this route for the penn data because the information was coming from two different record fields.

ndushay:
@jacobthill for the record, you could also pair with someone else on test and so on -- Aaron or Mike could help if you can get together with one of them sooner.

PR #251, opened 2019-10-07
date_parsing has replaced normalize_date

ndushay:
## Why was this change made?

Now that dlme-traject is using DateParsing.penn_museum_date_range macro for `cho_date_range_norm`, this is no longer used.

n/a

## Question

@jacobthill would you prefer that the DateParsing module be renamed NormalizeDate ?


mjgiarlo:
merging, but do see @ndushay's question above, @jacobthill

ndushay:
I'm happy to do a follow-on PR if desired.

PR #260, opened 2019-10-08
add missing requirement

jacobthill:
## Why was this change made?
Address missing requirement from record_counter.rb

n/a

### DIFF
--- a/lib/record_counter.rb
+++ b/lib/record_counter.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'concurrent'
+require 'singleton'
 
 module Dlme
   # A singleton counter for records.


lib/record_counter.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'concurrent'
+require 'singleton'
```
mjgiarlo: Ideally we would have a test as well to avoid regressions.

jacobthill:
@mjgiarlo can you add that test? I'm not sure how to do it.

mjgiarlo:
@jacobthill no problem, I'll toss a PR up that supersedes this one. thanks for finding and diagnosing it!

PR #268, opened 2019-10-09
Add hijri_range macro and cho_date_range_hijri

### LINKED ISSUE #189: Hijri to Gregorian and Gregorian to Hijri conversion
Need at macro to convert Hijri dates to Gregorian dates and Gregorian dates to Hijri dates. There is a Hijri date library that might be worth looking at: https://github.com/ecleel/hijri

aaron-collier:
## Why was this change made?

Fixes #189 

- Renames the `cho_date_hijri` field to `cho_date_range_hijri` to properly reflect the data.
- Adds `hijri_range` macro to `date_parsing` and a corresponding test
- Adds the proper `to_field` for `cho_date_range_hijri` to traject configs that currently have `cho_date_range_norm` configured.

N/A


aaron-collier:
@ndushay yes, we need a hijri based range facet. and yeah, `.to_hijri` is added by the hijri gem.

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,41 @@
     end
   end
 
+  describe '#hijri_range' do
+    it 'receives a range of integers' do
+      indexer.instance_eval do
+        to_field 'int_array', accumulate { |record, *_| record[:value] }, hijri_range
```
mjgiarlo: Since this appears to be the same in each test in this describe block, consider moving to a `before` block and out of the `it` statements?

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,41 @@
...
+      # note this is the first valid date that can be converted
+      expect(indexer.map_record(value: [623])).to include 'int_array' => [1, 2]
+    end
+
+    it 'is not provided a value' do
+      indexer.instance_eval do
+        to_field 'int_array', accumulate { |record, *_| record[:value] }, hijri_range
+      end
+
+      expect(indexer.map_record(value: [])).not_to include 'range'
```
mjgiarlo: what is `'range'` here? I was surprised not to see `'int_array'` here.

mjgiarlo: should this be `expect(indexer.map_record(value: [])).be_empty?` (assuming an empty array is returned)

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,41 @@
...
+    end
+
+    it 'receives an invalid value' do
+      indexer.instance_eval do
+        to_field 'int_array', accumulate { |record, *_| record[:value] }, hijri_range
+      end
+
+      expect(indexer.map_record(value: [622])).not_to include 'int_array' => [1, 2]
```
mjgiarlo: should this be `expect(indexer.map_record(value: [])).be_empty?` (assuming an empty array is returned)

spec/lib/traject/macros/date_parsing_spec.rb
aaron-collier: Happy to make this change, was following the pattern established in the spec, but clearing it up makes sense.

spec/lib/traject/macros/date_parsing_spec.rb
aaron-collier: `int_array` is just a generic place holder "field". I think you're right about `range` being wrong here.

lib/macros/date_parsing.rb
```diff
@@ -58,6 +59,21 @@ def fgdc_date_range
...
+    FIRST_VALID_GREGORIAN_TO_HIJRI_YEAR = 623
+
+    # Takes an existing array of year integers and returns an array converted to hijri
+    # with an additional year added to the end to account for the non-365 day calendar
+    def hijri_range
+      lambda do |_record, accumulator, _context|
+        if accumulator.any? { |val| val < FIRST_VALID_GREGORIAN_TO_HIJRI_YEAR } # 623 is the first valid year to convert
+          accumulator.replace([])
+          return
+        end
```
mjgiarlo: @aaron-collier Before merging, it'd be good to hear from @jacobthill what the right behavior is when a date range includes values outside (*i.e.*, before) the Hijri range? Should the macro:

1. return an empty result set (what's done above);
1. return an array filtering out these values; or
1. raise an exception

:question: cc: @ndushay

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,31 @@
     end
   end
 
+  describe '#hijri_range' do
+    before(:each) do
```
mjgiarlo: You can drop `(:each)` here and use `before do`, which defaults to the `:each` behavior. This is no biggie.

lib/macros/date_parsing.rb
ndushay: I have no idea.

spec/lib/traject/macros/date_parsing_spec.rb
ndushay: I've also done this:
```ruby
{
  'foo' => 'bar,
  'fickle' => 'hijri'
}.each_pair do | raw, expected|
   it "blah #{raw} blah #{expected" do
      expect(magic(raw)).to eq expected
   end
end
```

It's compact, and it's clear if you're checking a bunch of similar stuff.  when I get the marc_date_range macro done, it's what I'll be doing for it.

It does make debugging tests harder as the line numbers are a bit wonky, but if you use "format documentation" for the failing spec, you can see what is going on.

spec/lib/traject/macros/date_parsing_spec.rb
ndushay: It is so hard to remember to test your "not" expects thoroughly.

lib/macros/date_parsing.rb
mjgiarlo: @ndushay Yeah, it's a @jacobthill question, but I wanted to loop you in as DLSS Date Czar :laughing:

aaron-collier: I'm going to do a bit of work utilizing the formulas provide in wiki (https://en.wikipedia.org/wiki/Hijri_year#Formula) in order to account for BH (Before Hijri) dates:

```
AH = 1.030684 × (CE − 621.5643)

or

AH = (CE − 622) × 33 ÷ 32
```

@jacobthill and I will be pairing on this before stand up this morning.

jacobthill: I think this formula will work fine for all CE and BCE dates. Just replace CE with any positive (CE) or negative (BCE) number. There is a lot of complexity in translating between calendars. There are many slightly different formulas with various degrees of accuracy/complexity. I don't know enough to argue for one or another. This one seems fine for now. Essentially the implications are that with large negative years, if the formula is off slightly, the year will be miscalculated by a few years or decades with very large negative numbers. I don't really see this as a problem since the Muslim, and the Western, world doesn't really have a precise, well-establish system for dating large BCE numbers anyway.

mjgiarlo: Looking at this discussion of formulae and such, I am now wondering whether the `hijri` gem shouldn't be handling this for us. (Otherwise, what value is it providing? :laughing:)

aaron-collier: Yup @mjgiarlo after discussion, I'm 99% sure I'll be ditching the gem.

mjgiarlo: cooooool.

Gemfile
```diff
@@ -7,7 +7,6 @@ gem 'activesupport', '~> 5.2'
 gem 'config'
 gem 'dry-validation'
 gem 'faraday'
-gem 'hijri'
```
ndushay: this is a small reason why we should consider adding gems with the code.  I believe I did the exact same splitting of PRs recently, tho (for parse_date, I believe), so it's definitely a hindsight observation.

lib/contracts/cho.rb
```diff
@@ -10,7 +10,7 @@ class CHO < Dry::Validation::Contract
       optional(:cho_coverage).array(:str?)
       optional(:cho_creator).array(:str?)
       optional(:cho_date).array(:str?)
-      optional(:cho_date_hijri).array(:str?)
+      optional(:cho_date_range_hijri).array(:integer)
       optional(:cho_date_range_norm).array(:integer) # this is an array of integers to capture the years in a range
```
ndushay: the comment on line 14 should either move up to line 13 or go away, no?

lib/macros/date_parsing.rb
```diff
@@ -67,6 +88,19 @@ def penn_museum_date_range
       end
     end
 
+    HIJRI_MODIFIER = 1.030684
+    HIJRI_OFFSET = 621.5643
+
+    # @param [Integer] a single year to be converted
+    # @return [Integer] a converted integer year
+    # This method uses the first formulat provided here: https://en.wikipedia.org/wiki/Hijri_year#Formula
```
ndushay: typo: formulat

lib/macros/date_parsing.rb
```diff
@@ -67,6 +88,19 @@ def penn_museum_date_range
...
+    # This assumes rounding down, not up
```
ndushay: isn't "floor" obviously rounding down?  Is this comment necessary?

lib/macros/date_parsing.rb
```diff
@@ -58,6 +58,27 @@ def fgdc_date_range
       end
     end
 
+    FIRST_VALID_GREGORIAN_TO_HIJRI_YEAR = 623
+
+    # Takes an existing array of year integers and returns an array converted to hijri
+    # with an additional year added to the end to account for the non-365 day calendar
+    def hijri_range
+      lambda do |_record, accumulator, _context|
+        unless accumulator.is_a? Array
```
ndushay: i'm not confident, but I thought traject always has an array in the accumulator???  Is there a particular situation here that would make this not true?

If it is always an array, doesn't it make this:

```
accumulator.map! do |val|
  Macros::DateParsing.to_hijri(accumulator.first)..Macros::DateParsing.to_hijri(accumulator.last) + 1).to_a)
end 
```
?

mjgiarlo: Would you consider collapsing lines 67-75 with:

```ruby
if !accumulator.is_a?(Array) || accumulator.empty?
  accumulator.replace([])
  return
end
```

lib/macros/date_parsing.rb
```diff
@@ -58,6 +58,27 @@ def fgdc_date_range
...
+          accumulator.replace([])
+          return
+        end
+
+        if accumulator.empty?
+          accumulator.replace([])
+          return
+        end
```
ndushay: if you need to avoid my suggestion above, this block can become:

```
return if accumulator.empty?
```

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,31 @@
...
   end
 
+  describe '#hijri_range' do
+    before do
+      indexer.instance_eval do
+        to_field 'int_array', accumulate { |record, *_| record[:value] }, hijri_range
+      end
+    end
+
+    it 'receives a range of integers' do
```
ndushay: I don't mind this format, but if parse_date is large, the spec will become large.  Okay to leave as is, or to change to more compact approach in PR #271.

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -119,6 +119,31 @@
...
+      expect(indexer.map_record(value: [2010, 2011, 2012])).to include 'int_array' => [1431, 1432, 1433, 1434]
+    end
+
+    it 'receives a single value' do
+      # note this is the first valid date that can be converted
```
ndushay: It would be good to have a test demonstrating what happens if the value is earlier than first valid date.  Possibly throwing an exception would be warranted to call out the bad data??

aaron-collier: Actually, this comment is cruft, this now allows earlier dates.

lib/macros/date_parsing.rb
aaron-collier: Doh, I wanted this, but totally brain farted on it.

lib/macros/date_parsing.rb
aaron-collier: Yes, I just wanted to call attention that we're purposefully doing it. But if that just makes it messy, happy to drop.

lib/macros/date_parsing.rb
aaron-collier: I think you're right.

spec/lib/traject/macros/date_parsing_spec.rb
aaron-collier: I'm not quite getting what you mean. Can we pair after lunch?

PR #271, opened 2019-10-09
add marc_date_range macro

ndushay:
## Why was this change made?

`marc_date_range` macro is for populating `cho_date_range_norm` field from MARC data;  penn-marc and michigan provide marc data.

Note that I have a branch for penn/marc to use this macro and it ran ok (but I didn't check the output)

n/a


lib/macros/date_parsing.rb
```diff
@@ -56,6 +56,28 @@ def fgdc_date_range
...
+    # Extracts dates from slice of MARC 008 field
+    #  to_field "date_range", extract_marc("008[06-14]"), marc_date_range
+    # see https://www.loc.gov/marc/bibliographic/bd008a.html
+    # does NOT work for BC dates (or negative dates) - because MARC 008 isn't set up for that
+    def marc_date_range
+      lambda do |_record, accumulator, _context|
+        val = accumulator.first
+        date_type = val[0]
+        if date_type == 's'
+          first_year = ParseDate.earliest_year(val[1, 4])
```
mjgiarlo: Can you use range syntax here too like you do in the `elsif` block? Or use the comma syntax in both places? Doesn't matter much which you choose as long as it's consistent.

ndushay: done - good catch.

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -114,6 +114,40 @@
...
+      'q159u159u' => (1590..1599).to_a,
+      's1554    ' => [1554],
+      's15uu    ' => (1500..1599).to_a,
+      's193u    ' => (1930..1939).to_a,
+      's08uu    ' => (800..899).to_a
+    }.each_pair do |raw_val, expected|
+      it "#{raw_val} from 008[06-14] gets correct result" do
+        get_raw_val = lambda do |record, accumulator|
+          accumulator << record[:raw]
+        end
```
mjgiarlo: It looks like this code is repeated in both tests. Could it get moved into a `let(:get_raw_val) do ... end` right inside the `describe` block?

ndushay: for only 2 test blocks, I didn't think it worth it.  Also, I have bumped into a lot of "you can't do this outside of an it block" and such, as well as scoping issues for these tests. 

Sandi Metz:  2 repeats - hold your nose.  3 repeats - really hold your nose.   4 repeats - refactor.

ndushay: I have had second thoughts about my response - DRY-it-up commit coming imminently

ndushay: and ... done.

spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -114,6 +114,37 @@
     end
   end
 
+  describe '#marc_date_range' do
+    get_raw_val = lambda do |record, accumulator|
+      accumulator << record[:raw]
+    end
```
mjgiarlo: Would you consider replacing this with:

```ruby
let(:get_raw_val) do
  lambda do |record, accumulator|
    accumulator << record[:raw]
  end
end
```

That looks more rspec-y than having a variable declaration in a `describe` block.

ndushay: sure!  wow, picky!

ndushay:
Mike, while I appreciate all I'm learning here, I'm wondering about the time spent and Sandi Metz "you do not have a god given right to create perfect code."

PR #275, opened 2019-10-10
Add array_from_range macro

aaron-collier:
## Why was this change made?

This adds a macro that will take the range dates available in AUC (and other collections) as "yyyy; yyyy; yyyy;..." and return an array.

N/A

### DIFF
--- a/lib/macros/date_parsing.rb
+++ b/lib/macros/date_parsing.rb
@@ -16,7 +16,7 @@ module Macros
 
         range_years = accumulator.first.delete(' ')
 
-        unless range_years.=~(/^[0-9-;]+$/)
+        unless range_years.match?(/^[0-9-;]+$/)
           accumulator.replace([])
           return
         end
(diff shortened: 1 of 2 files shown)


lib/macros/date_parsing.rb
```diff
@@ -7,6 +7,24 @@
...
+    # get array of year values in range, when string is:
+    # yyyy; yyyy; yyyy; yyyy; yyyy
+    # works with negative years, but will return an emtpy set of a string is detected
+    def array_from_range
+      lambda do |_record, accumulator|
+        return if accumulator.empty?
+
+        range_years = accumulator.first.delete(' ')
+
+        unless range_years.=~(/^[0-9-;]+$/)
```
ndushay: i haven't seen this notation (with the dot) before.  it's a little ... compact?  Wow.

aaron-collier: Good catch, it works.. but it should really be without the dot and the parens, so pushed up a quick change.

PR #280, opened 2019-10-11
American numismatic date parsing

jacobthill:
## Why was this change made?

- Adds macro to parse American Numismatic Society dates
- Calls macro in numismatic_config.rb
- Adds tests for macro

n/a

### DIFF
--- a/spec/lib/traject/macros/date_parsing_spec.rb
+++ b/spec/lib/traject/macros/date_parsing_spec.rb
@@ -36,6 +36,12 @@ RSpec.describe Macros::DateParsing do
     end
   end
 
+  let(:raw_val_lambda) do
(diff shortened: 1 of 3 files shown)


spec/lib/traject/macros/date_parsing_spec.rb
```diff
@@ -12,6 +12,10 @@
     end
   end
 
+  let(:raw_val_lambda) do
+    lambda do |record, accumulator|
+      accumulator << record[:raw]
+
```
mjgiarlo: You're missing a couple `end` statements here: one to terminate the `lambda do` and one to terminate the `let() do`.

jacobthill:
Not sure what to do about this error.

PR #287, opened 2019-10-15
add Arabic translation map for cho_has_type

### LINKED ISSUE #155: Normalize incoming types to cho_has_type
There are three type values in the dlme [application_profile](https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md): cho_edm_type, cho_type, and cho_has_type. The first is implemented, the second is used to store type values passed through from the incoming data without alteration, the third (cho_has_type) needs to be implemented as sub categories of cho_edm_type which captures the type of media (e.g. records with cho_edm_type=Image can be assigned cho_has_type values of 'postcard', 'coin', 'poster', 'painting', 'photograph', etc.)

To do before writing requirements: 
- [ ] create Arabic equivalent of translation maps

jacobthill:
## Why was this change made?
Add Arabic translation map for cho_has_type values, closes #155. An additional translation map may be necessary but this won't be clear until the mapping is underway.

n/a

### DIFF
--- a/lib/translation_maps/norm_has_type_to_ar.yaml
+++ b/lib/translation_maps/norm_has_type_to_ar.yaml
@@ -1,4 +1,3 @@
-
 Cultural Artifact: قطعة أثرية ثقافية
 Drawing: رسم
 Map: خريطة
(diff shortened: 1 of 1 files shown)


lib/translation_maps/norm_has_type_to_ar.yaml
```diff
@@ -0,0 +1,16 @@
+
```
mjgiarlo: Is the blank line here intentional? Dunno if it's problematic but thought I'd ask first!

PR #296, opened 2019-10-22
add missing date range fields we can currently process to traject configs

ndushay:
## Why was this change made?

This doc shows the status of current DLME collections for date range normalization

https://docs.google.com/spreadsheets/d/1U068CQlySkxiS2O5W70Uc4uc6t1DrqJGMcUO71CLokY/edit#gid=0

the following collections are supposedly "done" but were missing the config lines.

not sure why some of these lost their `cho_date_range_norm` stuff.  (perhaps a rebase conflict in some later PR, but it doesn't matter.  And `cho_date_range_norm` being missing probably impacted `cho_date_range_hijri` not being added, but again it doesn't matter.)

n/a


ndushay:
merged my own here, with one approval.

PR #298, opened 2019-10-24
Refactor tweaks

ndushay:
## Why was this change made?

To get rid of an annoying warning / possibly incorrect regex, and to have better comments for 2 of the date parsing macros.

n/a

### DIFF
--- a/lib/macros/date_parsing.rb
+++ b/lib/macros/date_parsing.rb
@@ -24,8 +24,9 @@ module Macros
       end
     end
 
-    # given a string with date info, use parse_date gem to get an array of indicated years as integers
-    #  See https://github.com/sul-dlss/parse_date for info on what it can parse
+    # given an accumulator containing a string with date info,
+    #   use parse_date gem to get an array of indicated years as integers
+    #   See https://github.com/sul-dlss/parse_date for info on what it can parse
     def parse_range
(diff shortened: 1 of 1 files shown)


lib/macros/date_parsing.rb
```diff
@@ -15,7 +15,7 @@ def array_from_range
 
         range_years = accumulator.first.delete(' ')
 
-        unless range_years.match?(/^[0-9-;]+$/)
+        unless range_years.match?(/^[0-9\-;]+$/)
```
ndushay: this line was giving a warning because the hyphen, when intended to be the hyphen char, was not regex escaped.

lib/macros/date_parsing.rb
```diff
@@ -58,7 +59,7 @@ def american_numismatic_date_range
     FGDC_DATE_RANGE_XPATH = "#{FGDC_TIMEINFO_XPATH}/rngdates"
     # Note:  saw no "#{FGDC_TIMEINFO_XPATH}/mdattim" multiple dates path data
 
-    # Extracts dates from FGDC idinfo/timeperd to create a singe date range value
+    # Extracts dates from FGDC idinfo/timeperd to create a single date range value
```
ndushay: i hate singe dates.

Issue #244, opened 2019-10-04
Normalize OPenn dates for date slider (cho_date_range_norm)


ndushay:
To populate cho_date_range_norm for Penn openn collections:

Date is in  `<tei:origDate>`, and unlike Cambridge, there are no attributes present to help EXCEPT for the single record for philadelphia art museum.  There are a variety of string parsing patterns I am going to try to group into patterns ...

Not yet sure if separate collection approach is best or not.

See issue #282 for new and exciting date string patterns to parse.

ndushay:
### Columbia subcoll

no hijri dates provided

### Free Library Philadelphia subcoll

some of the date strings have A.H. and Gregorian

patterns found:
```
A.H. 986 (1578)
A.H. 1240 (1824)
1269 A.H. (1852)
```

### penn-0001 subcoll

some of the date strings have A.H. and Gregorian

patterns found:
```
A.H. 986 (1578)
A.H. 899 (1493-1494)
A.H. 901-904 (1496-1499)
A.H. 1240 (1824)
ca. 1670 (A.H. 1081)
```

### penn-0002 subcoll

some of the date strings have A.H. and Gregorian

patterns found:
```
A.H. 986 (1578)
A.H. 1240 (1824)
A.H. 1258? (1842)
A.H. 1224, 1259 (1809, 1843)
A.H. 1123?-1225 (1711?-1810)
```

### philadelphia art museum subcoll

one record with TEI "notBefore" and "notAfter" attributes

```xml
        <history>
            <origin>
              <origDate notBefore="1400" notAfter="1599"/>
              <p>Copied in India or elsewhere in the Mughal Empire between the 14th and 16th centuries. No date and no scribal attribution; between 1400 and 1599</p>
              <origPlace>[India?]</origPlace>
            </origin>
```

ndushay:
PR #301 does all but philadelphia art museum, which will need special treatment.  (and ... for ONE record???)

PR #301, opened 2019-10-25
4 OPenn colls get date ranges

ndushay:
## Why was this change made?

Provides date ranges for 4 (out of 5) OPenn collections - most of #244.

n/a

## Notes

Once again I have proved that I shouldn't do the macros as a separate PR from hooking up the collections to use the macro.  🙄


config/metadata_mapping.json
```diff
@@ -406,7 +409,7 @@
   },
   {
     "trajects": [
-      "openn_config.rb"
+      "openn_common_config.rb"
```
ndushay: openn/philadelphia-museum-art will get need to get start and end dates for date range from attributes on `<origDate>`

```
        <history>
            <origin>
              <origDate notBefore="1400" notAfter="1599"/>
              <p>Copied in India or elsewhere in the Mughal Empire between the 14th and 16th centuries. No date and no scribal attribution; between 1400 and 1599</p>
              <origPlace>[India?]</origPlace>
            </origin>
```

lib/macros/date_parsing.rb
```diff
@@ -31,7 +31,7 @@ def parse_range
       lambda do |_record, accumulator|
         range_years = []
         accumulator.each do |val|
-          range_years << ParseDate.parse_range(val)
+          range_years << ParseDate.parse_range(val) if val&.strip.present?
```
ndushay: I believe we had some empty dates in openn, so this guard clause became necessary.

lib/macros/date_parsing.rb
```diff
@@ -53,27 +54,39 @@ def hijri_from_mixed(date_str)
     end
 
     # given an accumulator containing a string with both hijri and gregorian date info,
-    #   change the accumulator contents to become only the hijri date info
-    def parse_hijri
+    #   change the string to only contain gregorian date info
+    def extract_gregorian
```
ndushay: `extract_gregorian` is the same - the diff is confused.  The comment did change a bit.

lib/macros/date_parsing.rb
```diff
@@ -53,27 +54,39 @@ def hijri_from_mixed(date_str)
...
       end
     end
 
-    # given an accumulator containing a string with both hijri and gregorian date info,
-    #   change the accumulator contents to become only the gregorian date info
-    def parse_gregorian
+    # given an accumulator containing a string that may (or may not) have both hijri and gregorian date info,
+    #   get a hijri range, either directly from the provided hijri dates or converted from the gregorian if no
+    #   hijri dates are provided.
+    def extract_or_compute_hijri_range
```
ndushay: I assumed it is better to take provided hijri years for `cho_date_range_hijri`, when we have them.

mjgiarlo: would this benefit from feedback from @jacobthill?

aaron-collier: Agreed, I'd like to see @jacobthill's feedback on the path here.

config/metadata_mapping.json
ndushay: @mjgiarlo this ^^ single-record collection is why the date parsing for the other 4 collections is in a file `openn_no_orig_date_attribs.rb` rather than in openn_common_config.rb.

ndushay:
@mjgiarlo openn/philadelphia-museum-art collection has attributes with dates we would want to use so it is a diff animal.  Hence the date fields for these 4 collections in their own file.

I'm wondering at the special snowflake status of the one record philly museum collection myself -- I took the data as I found it.

config/metadata_mapping.json
mjgiarlo: Aha. Thanks! I didn't make the connection.

PR #306, opened 2019-10-30
Fix multilingual bug

mjgiarlo:
## Why was this change made?

Fixes sul-dlss/dlme#603

This bug was caused by having a Traject config (`stanford_mods_config.rb`) attempt to convert a field (`cho_title`) to the language hash which was already converted to the language hash in an earlier config (`mods_config.rb`). I added a validation that will fail the transformation if we do this again (a bug I introduced :D)

No.

### DIFF
--- a/lib/contracts/cho.rb
+++ b/lib/contracts/cho.rb
@@ -67,14 +67,19 @@ module Contracts
       end
     end
 
+    # rubocop:disable Metrics/AbcSize
     def self.required_language_specific_rule
       proc do
         key.failure('no values provided') if value.keys.empty?
         unexpected_keys = value.keys - expected_language_values
         key.failure("unexpected language code(s) found in #{key.path.keys.first}: #{unexpected_keys.join(', ')}") if
           unexpected_keys.any?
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@ndushay :speech_balloon: 
> I wonder if we should default the cho_title to be english for all the collections.

IIRC, I asked @jacobthill about this and he was against it for reasons that I can't remember. Why do you think we should do that? Having `none` values isn't causing the bug fixed by :point_up: and matches the spec we were asked to implement.

spec/integration/traject_configs_spec.rb
```diff
@@ -19,6 +19,6 @@
           settings: config.fetch('settings', {})
         ).transform
       end
-    end.not_to raise_error(RuntimeError)
+    end.not_to raise_error
```
ndushay: I'm also wondering if we can have a better test ... maybe for stanford_mods macros ... that shows that the title value is correct.

mjgiarlo:
@ndushay can you take another look?

PR #307, opened 2019-10-30
Check whether *any* exception is raised in integration spec

mjgiarlo:
## Why was this change made?

Only checking for `RuntimeError` allows for false positives, where we may be missing exceptions. 

This is currently failing because of a data error, and indeed, there could be several dozen or hundred more such errors. As we fix data that does not cleanly map, we should revisit this PR until we get a clean build. 

no.


mjgiarlo:
Blocked by https://github.com/sul-dlss/dlme-metadata/pull/123

mjgiarlo:
This will be superseded by https://github.com/sul-dlss/dlme-transform/issues/325. Closing.

PR #310, opened 2019-10-31
update cho_has_type translations

jacobthill:
## Why was this change made?

To update the cho_has_type translations based on feedback from QNL

n/a


mjgiarlo:
Per @aaron-collier, PRs like this are still in @sul-dlss/infrastructure-team's bailiwick, so I'm merging. Thanks, @ndushay and @jacobthill

PR #314, opened 2019-11-01
Remove config from docker ignore as it is now required

aaron-collier:
## Why was this change made?
We were previously pulling configs from dlme-traject, we know need the base configs from this repository and with them ignored (and thus not built into the container) an error is thrown during the run:

```
[ERROR] No such file or directory @ rb_sysopen - /opt/traject/config/metadata_mapping.json
/opt/traject/lib/config_finder.rb:67:in `read': No such file or directory @ rb_sysopen - /opt/traject/config/metadata_mapping.json (Errno::ENOENT)
```
Note: this is blocking transforms from running in docker and AWS, but not if running manually.

N/A

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
 /output/*
 /data/*
-/config/*


aaron-collier:
Confirmed I can now run transforms in AWS Dev.

mjgiarlo:
Ooh, good catch, @aaron-collier

Issue #326, opened 2019-11-04
re-analyze AU Cairo collections for date range parsing


ndushay:
#### single `dc:date` field
1. coll 1 - all same value:  `1960; 1961; ... ; 1989` - `parse_range` macro will work
2. coll 2 - `parse_range` macro will work
3. coll 5 - `yyyy`, `undated`, `yyyy; yyyy; yyyy` - existing auc macro will work
4. coll 9 - all same value:  `1882` - `parse_range` macro will work
5. coll 11 - `yyyy`, `yyyy; yyyy; yyyy` - existing auc macro will work
6. coll 12 - all same value:  `1980` - `parse_range` macro will work
7. coll 15 - `yyyy`, `yyyy-mm`, `[n.d. 1914?]` - `parse_range` macro will work
8. coll 16 - all same value:  `1870s`, `parse_range` macro will work eventually
9. coll 19 - `yyyy-mm` - `parse_range` macro will work
10. coll 24 - `2016-mm-dd` (yes, valid per Jacob) - `parse_range` macro will work
11. coll 26 - `yyyy-mm-dd`, `yyyy-mm` - `parse_range` macro will work
12. coll 29 - `yyyy-mm-dd` - `parse_range` macro will work
13. coll 30 - existing auc macro will work (one semicolon, rest ok for `parse_range`; yes Gregorian per Jacob)
14. coll 31  - all same value: `1964-1967?` - `parse_range` macro will work

#### multiple `dc:date` fields
- coll 4 - existing auc macro will work, using all `dc:date` occurrences
- coll 6 - existing auc macro will work, using all `dc:date` occurrences
- coll 8 - TODO: ignore all but first `dc:date` field
- coll 13 - existing auc macro will work, using all `dc:date` occurrences
- coll 14 - all same values `ca. late 19th century or early 20th century` + `yyyy; yyyy; ...` - existing auc macro (take range if provided) should work
- coll 21 - as coll 14, but a few `yyyy` values thrown in, too
- coll 23 - existing auc macro (take range if provided) should work
- coll 32 - existing auc macro should work

ndushay:
Given above, my suggestion is:

We have `auc_common_config`
We have `auc_date_range_from_all_config` (or some better name) and use it for getting date range fields for all colls except coll 8.   It uses the existing auc date parsing macro for the date range fields, utilizing all occurences of `dc:date`.

[...]

PR #327, opened 2019-11-05
auc coll 8 only uses first dc:date for date range; rest of auc u…

ndushay:
DO NOT MERGE without Jacob's approval

## Why was this change made?

Related to #326 

Some AU Cairo dates in the date range fields were incorrect.   While in some ways this is caused by bad data, Jacob felt the recent years added to coll 8 were particularly egregious.

This PR is a way to get better values for coll 8 while hopefully retaining excellent values for all the other collections.

n/a ???

## NOTE

@jacobthill must sign off on this approach.  Do not merge to master without his approval.


ndushay:
See Jacob's comment on issue #326 -- so far, still problems.

ndushay:
@jacobthill I pushed up a fix that addresses all the problems you reported

PR #349, opened 2019-11-06
parse_date v0.4.2 get correct date ranges for yyyy-y pattern

ndushay:
## Why was this change made?

See https://github.com/sul-dlss/parse_date/issues/33 -- at least 3 collections have some dates with yyyy-y pattern;  this will ensure we get correct range of years.

n/a

### DIFF
--- a/traject_configs/bodleian_config.rb
+++ b/traject_configs/bodleian_config.rb
@@ -28,8 +28,8 @@ to_field 'cho_title', extract_json('.title'), strip, default('Untitled Item')
 to_field 'cho_creator', extract_json('.author'), strip
 to_field 'cho_contributor', extract_json('.printer'), strip, append(' [printer]')
 to_field 'cho_date', extract_json('.date_statement'), strip
-to_field 'cho_date_range_norm', extract_json('.date_statement'), strip, parse_range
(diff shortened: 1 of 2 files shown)


traject_configs/bodleian_config.rb
```diff
@@ -28,8 +28,8 @@
 to_field 'cho_creator', extract_json('.author'), strip
 to_field 'cho_contributor', extract_json('.printer'), strip, append(' [printer]')
 to_field 'cho_date', extract_json('.date_statement'), strip
-to_field 'cho_date_range_norm', extract_json('.date_statement'), strip, parse_range
-to_field 'cho_date_range_hijri', extract_json('.date_statement'), strip, parse_range, hijri_range
+to_field 'cho_date_range_norm', extract_json('.date_statement'), strip, gsub('/', '-'), parse_range
+to_field 'cho_date_range_hijri', extract_json('.date_statement'), strip, gsub('/', '-'), parse_range, hijri_range
```
ndushay: Armenian collection has `1974/5`;  there are no slashes anywhere else so this improves Armenian coll date range results without negatively impacting other date range results.

PR #358, opened 2019-11-07
Revise AIMS transform to add language values, fixes #348

### LINKED ISSUE #348: Check, revise, and transform AIMS
After significant changes to the IR, it is necessary to check, possibly revise, re-transform, and re-load the AIMS data set

jacobthill:
## Why was this change made?
Needed to revise AIMS mapping to reflect changes to the IR and add language keys.

n/a


traject_configs/aims_config.rb
```diff
@@ -1,17 +1,17 @@
...
+require 'macros/dlme'
 require 'macros/post_process'
+require 'traject_plus'
 
-extend Macros::PostProcess
-extend Macros::DLME
+extend Macros::AIMS
 extend Macros::DateParsing
+extend Macros::DLME
 extend TrajectPlus::Macros
```
ndushay: if these are alphabetical, this line should be last.

PR #372, opened 2019-11-07
Revise Bodleian and Princeton mappings

jacobthill:
## Why was this change made?

- Revise bodleian mapping

- add cho_has_type and cho_identifier fields

n/a


traject_configs/princeton_mss_config.rb
```diff
@@ -18,53 +18,52 @@
 end
 
 # Cho Required
-to_field 'id', extract_json('.identifier'), strip
```
ndushay: is strip removed because it's not needed?   Is it causing harm?   Just curious.

PR #375, opened 2019-11-08
Do not raise exceptions for any date-parsing errors

### LINKED ISSUE #325: Catch exception where first year > last year
The parse_date gem raises an error in this case, but for AUC (and likely other collection) we want to capture this error and log it as a warning in order to fix at the provider.

mjgiarlo:
Fixes #325

## Why was this change made?

Collect date-parsing errors as transformations run and then log them at the end of the run. This way, @jacobthill can see more clearly which datasets need changes, without bombing a whole batch.

Also includes:
* Bump `parse_date` to 0.4.2 so that year range exceptions come back as `ParseDate::Error` rather than `StandardError`.
* Rename `PostProcess` macros to `EachRecord` to distinguish them from the `after_processing` code
* Let integration specs fail if any exceptions are raised so we can find errors sooner as we go forward.

No.

### DIFF
--- a/traject_configs/aims_config.rb
+++ b/traject_configs/aims_config.rb
@@ -4,13 +4,13 @@ require 'dlme_json_resource_writer'
 require 'macros/aims'
 require 'macros/date_parsing'
 require 'macros/dlme'
-require 'macros/post_process'
+require 'macros/each_record'
 require 'traject_plus'
 
 extend Macros::AIMS
-extend Macros::DateParsing
 extend Macros::DLME
-extend Macros::PostProcess
+extend Macros::DateParsing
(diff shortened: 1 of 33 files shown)


traject_configs/aims_config.rb
```diff
@@ -4,13 +4,13 @@
 require 'macros/aims'
 require 'macros/date_parsing'
 require 'macros/dlme'
-require 'macros/post_process'
+require 'macros/each_record'
```
ndushay: so these should now be alphabetical.  (I know, I know - it's Jacob as much as me). requires and extends ...

mjgiarlo: done

PR #376, opened 2019-11-08
Revise ANS mapping, fixes #347

### LINKED ISSUE #347: Check, revise, and transform American Numismatic
After significant changes to the IR, it is necessary to check, possibly revise, re-transform, and re-load the American Numismatic data set.

jacobthill:
## Why was this change made?

- Revise ANS mapping

- Update metadata-mappings

n/a

### DIFF
--- a/config/metadata_mapping.json
+++ b/config/metadata_mapping.json
@@ -24,7 +24,7 @@
       "numismatic_csv_config.rb"
     ],
     "paths": [
-      "american-numismatic-society/islamic-department"
+      "american-numismatic-society"
     ],
     "extension": ".csv",
     "settings": {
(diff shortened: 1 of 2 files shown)


aaron-collier:
Same question(s). This LGTM, but since `cho_has_type` is the second level of the hierarchical facet, should it be here or do we just not get/have the data for it?

jacobthill:
I can only map fields that are available in the metadata or clear from observation. There are no cho_language, cho_has_type, or cho_has_part values in the metadata. cho_format was mipmapped; I remapped the 'Object Type' column to 'cho_type'. That seemed more appropriate to me.

PR #380, opened 2019-11-11
Revise Sakip Sabanci mapping, fixes #356

### LINKED ISSUE #356: Check, revise, and transform Sakip Sabanci
After significant changes to the IR, it is necessary to check, possibly revise, re-transform, and re-load the Sakip Sabanci data sets.

jacobthill:
## Why was this change made?

- revise Sakip Sabanci mapping

n/a

### DIFF
--- a/traject_configs/sakip_sabanci_common_config.rb
+++ b/traject_configs/sakip_sabanci_common_config.rb
@@ -5,45 +5,47 @@ require 'macros/content_dm'
 require 'macros/date_parsing'
 require 'macros/dlme'
 require 'macros/each_record'
+require 'macros/normalize_language'
 require 'macros/normalize_type'
 require 'macros/oai'
 require 'traject_plus'
 
 extend Macros::ContentDm
-extend Macros::DLME
 extend Macros::DateParsing
+extend Macros::DLME
 extend Macros::EachRecord
+extend Macros::NormalizeLanguage
 extend Macros::NormalizeType
 extend Macros::OAI
 extend TrajectPlus::Macros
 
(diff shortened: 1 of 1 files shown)


jacobthill:
@ndushay I haven't figured out a way to map the cho_has_type field yet. This can only be mapped when the collections are somewhat homogenous. I have an open ticket to translate the Turkish types to English and then use existing translation maps to map the English to Arabic. When I work on that I will try to see if I can map the cho_has_type field but I think it can be merged for now.

ndushay:
@jacobthill lang code for ?? turkish ?? incorrect, as circleci says:

```
<RuntimeError: error raised mapping sakip-sabanci: Traject::Indexer::ConfigLoadError: Error loading ..._configs/sakip_sabanci_common_config.rb:29 RuntimeError:tr is not an acceptable BCP47 language code>
```

jacobthill:
@mjgiarlo Do you know why we are getting this error. I believe `tr` is the correct language code for Turkish. I think we did something to call this error for all language fields that were not explicitly validated by us (like by adding them in a translation map or something) but I can't find where we might have done that. I thought I asked you to do this so I wouldn't pass the wrong language field by mistake without any warning. Does this sound familiar at all?

mjgiarlo:
@jacobthill It's because `tr` is not in this list: https://github.com/sul-dlss/dlme-transform/blob/master/config/settings.yml#L13-L21

mjgiarlo:
Blocked by https://github.com/sul-dlss/dlme-transform/pull/395

PR #386, opened 2019-11-13
 add_cho_type_facet  - new each_record macro for hierarchical facet field

### LINKED ISSUE #383: Add field to support hierarchical facet display
In order to support the hierarchical facet described in sul-dlss/dlme#552 we need a field that has the the entire hierarchy expanded and delimited with a consistent value (that will not be in the data). You can see an example of the expected solr fields to support the hierarchy in the [Usage section](https://github.com/sul-dlss/blacklight-hierarchy#usage) of the blacklight-hierarchy gem.

I'm not 100% sure which two fields are intended to be included in the hierarchy.  I see `cho_edm_type_ssim` (the current Type facet) and `cho_type_ssim` (which is listed as the Other type facet, but I haven't seen it in any data yet).  So this might be something that needs a bit of clarification if we don't have the data to support this display yet.

ndushay:
## Why was this change made?

Closes #383 - Access team needs field to support hierarchical facet display 

I confirmed values with Jessie and Aaron.  This may not be perfect, but it is hopefully enough for Access to be unblocked on this.

After this is merged, we should push a new docker image and redeploy to ... dev?

n/a -- I can't think of a place appropriate to document this.

### DIFF
--- a/lib/macros/each_record.rb
+++ b/lib/macros/each_record.rb
@@ -3,6 +3,65 @@
 module Macros
   # Macros for post-processing data
   module EachRecord
+    # NOTE: compute cho_type_facet BEFORE calling convert_to_language_hash fields
+    # NOTE: do *not* include cho_type_facet in convert_to_language_hash fields
+    # create cho_type_facet values from cho_edm_type and cho_has_type values,
+    # add to context.output_hash if there's a value
+    def add_cho_type_facet
+      lambda do |_record, context|
+        context.output_hash['cho_type_facet'] = {}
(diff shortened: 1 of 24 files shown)


mjgiarlo:
@ndushay :speech_balloon: 
> 
> After this is merged, we should push a new docker image and redeploy to ... dev?
> 

CircleCI should take care of building and pushing a new docker image for us, and this latter bit is effectively how we "deploy" this codebase so... once we merge, we should be done. Confirm/deny, @aaron-collier ?

PR #391, opened 2019-11-15
Return first value in addition to joined value for hierarchical facet values

### LINKED ISSUE #390: Update cho_type_facet to add top level item by itself
It looks like w/o the top level node/type as the first item in the field it's not linked (which I guess is what it says in the [README](https://github.com/sul-dlss/blacklight-hierarchy#usage), so that shouldn't surprise me).

So for a field that was:
```ruby
['Sound:Interview']
```
We would like the `Sound` value repeated as the first value in the field

```ruby
['Sound', 'Sound:Interview']
```

This will allow the Blacklight Hierarchy gem to link both the top level type as well as the sub type to be filtered on.

mjgiarlo:
Fixes #390

## Why was this change made?

To feed `blacklight-hierarchy` the values it expects.

no.

### DIFF
--- a/lib/macros/each_record.rb
+++ b/lib/macros/each_record.rb
@@ -40,7 +40,7 @@ module Macros
           edm_type_first = values&.first
           has_type_first = cho_has_type_raw_hash[lang]&.first if cho_has_type_raw_hash
           val = hierarchical_val(edm_type_first, has_type_first)
-          context.output_hash['cho_type_facet'][lang] = [val] if val.present?
+          context.output_hash['cho_type_facet'][lang] = val if val.present?
         end
         context.output_hash.delete('cho_type_facet') if context.output_hash['cho_type_facet'].empty?
       end
@@ -51,7 +51,10 @@ module Macros
     # helper method for add_cho_type_facet
     # @return [String or nil]
     def hierarchical_val(*values)
-      values.compact.join(HIER_LEVEL_SEP_CHAR)
+      non_nil_values = values.compact
+      return non_nil_values if non_nil_values.count < 2
+
+      [non_nil_values.first, non_nil_values.join(HIER_LEVEL_SEP_CHAR)]
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/macros/each_record.rb
```diff
@@ -40,7 +40,7 @@ def add_cho_type_facet
           edm_type_first = values&.first
           has_type_first = cho_has_type_raw_hash[lang]&.first if cho_has_type_raw_hash
           val = hierarchical_val(edm_type_first, has_type_first)
-          context.output_hash['cho_type_facet'][lang] = [val] if val.present?
+          context.output_hash['cho_type_facet'][lang] = val if val.present?
```
mjgiarlo: `hierarchical_val` always returns an array now, so don't wrap in another array.

lib/macros/each_record.rb
```diff
@@ -51,7 +51,10 @@ def add_cho_type_facet
     # helper method for add_cho_type_facet
     # @return [String or nil]
     def hierarchical_val(*values)
-      values.compact.join(HIER_LEVEL_SEP_CHAR)
+      non_nil_values = values.compact
+      return non_nil_values if non_nil_values.count < 2
```
mjgiarlo: Nothing to join, move on.

mjgiarlo:
@jkeck :point_up:

jkeck:
thanks!

mjgiarlo:
i gotchu

lib/macros/each_record.rb
ndushay: I'm not sure this will work if there is no value for `cho_has_type`.  I'm not sure if it's supposed to return `[Sound]` on its own or `[Sound, Sound]` but perhaps you already inquired about that.

jkeck: I would expect just `[Sound]` in the case there was no sub-type.

mjgiarlo: @jkeck :point_up: :question:

ndushay: excellent.

mjgiarlo:
Hey, @jkeck! Re:
> 
> Also, after this is merged, would you please deploy to dev (I think bug fix increment is good for tagged release) and ask Jacob to delete and then index AIMS and PPPA again? 

dlme-transform 1.7.1 has been tagged, and I've created the terraform-aws PR: https://github.com/sul-dlss/terraform-aws/pull/526

PR #395, opened 2019-11-19
Allow Turkish language without script designation

mjgiarlo:
## Why was this change made?

Requested by @jacobthill 

no

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -17,8 +17,7 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
-  - tr-Arab
-  - tr-Latn
+  - tr
 defaults:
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -17,8 +17,7 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
-  - tr-Arab
-  - tr-Latn
+  - tr
```
aaron-collier: should we make this change or all (i.e. fa-Arab/fa-Latn above)? Or no?

mjgiarlo: @aaron-collier I believe @jacobthill only wanted a change to how we handle Turkish.

jacobthill:
Looks good, thanks @mjgiarlo

PR #396, opened 2019-11-19
Remap Penn museum

jacobthill:
## Why was this change made?

- revise penn config

 n/a

### DIFF
--- a/lib/macros/penn.rb
+++ b/lib/macros/penn.rb
@@ -9,12 +9,12 @@ module Macros
     # @return [Proc] a proc that traject can call for each record
     def penn_cho_has_type
       lambda do |_record, accumulator|
-        if accumulator.first == 'Manuscript Fragment'
-          accumulator.first.gsub('Manuscript Fragment', 'MS WORKED')
-        elsif accumulator.first == 'Manuscript'
-          accumulator.first.gsub('Manuscript', 'Manuscript ALSO WORKED')
+        if accumulator[0]&.match('Manuscript Frangment')
+          accumulator.replace(['Manuscript'])
+        elsif accumulator[0]&.match('Manuscript')
+          accumulator.replace(['Manuscript'])
         else
-          accumulator.first.gsub(/./, 'Cultural Artifact')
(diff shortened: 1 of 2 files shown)


jacobthill:
@mjgiarlo, @aaron-collier this is probably super easy but how can I assign values to the accumulator in the penn.rb file in this PR? I want to replace 'Manuscript Fragment' with 'Manuscript', leave 'Manuscript' values as is, and replace anything else with 'Cultural Artifact' but I can't figure out how to overwrite the value. In python its something like: 

```
if accumulator == 'Manuscript Fragment':
    accumulator = 'Manuscript'
```
How do I do the equivalent in ruby?

lib/macros/penn.rb
```diff
@@ -0,0 +1,22 @@
...
+    #   penn_cho_has_type => lambda { ... }
+    # @return [Proc] a proc that traject can call for each record
+    def penn_cho_has_type
+      lambda do |_record, accumulator|
+        if accumulator.first == 'Manuscript Fragment'
+          accumulator.first.gsub('Manuscript Fragment', 'MS WORKED')
+        elsif accumulator.first == 'Manuscript'
+          accumulator.first.gsub('Manuscript', 'Manuscript ALSO WORKED')
+        else
+          accumulator.first.gsub(/./, 'Cultural Artifact')
```
mjgiarlo: @jacobthill is this the line that's not doing what you expect? Give `accumulator.replace(['Cultural Artifact'])` a try.

jacobthill: Thanks @mjgiarlo that works except I can't get the if condition to match so it replaces everything with ['Cultural Artifact']. How do I check to see if the accumulator is ['Manuscript'] etc. before overwriting it? I've tried
```
if accumulator[0]&.match(['Manuscript Frangment'])
```
and 
```
if accumulator[0]&.match('Manuscript Frangment')
```
and
```
if accumulator.first&.match('Manuscript Frangment')
```
etc.

jacobthill: Nevermind, I got it. thanks

PR #403, opened 2019-11-20
Revise Princeton mapping, fixes #354

### LINKED ISSUE #354: Check, revise, and transform Princeton
After significant changes to the IR, it is necessary to check, possibly revise, re-transform, and re-load the Princeton data sets.

jacobthill:
## Why was this change made?

- Revise Princeton mapping

- n/a


jacobthill:
@aaron-collier, @mjgiarlo This error seems to be related to the start-date being later than the end-date which I thought was fixed. Any idea how to resolve this?

traject_configs/princeton_yemeni_mss_config.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Cho Other
```
mjgiarlo: @jacobthill I believe one of the errors in this branch may be due to using the `lang()` macro without having required/extended the DLME macros module in this config. cc: @aaron-collier

jacobthill: @mjgiarlo that didn't seem to work. Also this is a typical pattern. Whenever I have a common config with collection specific configs as well, all the requirements/enxtensions are in the common but not repeated in the collection specific configs (e.g. bnf, michigan, etc.). Also this seems to be complaining about a michigan record though nothing in this PR should effect michigan (until I just added the extensions). Am I missing something?

jacobthill: Nevermind, I figured it out. There were two problems (with michigan and penn_marc).

mjgiarlo: Good find!

Issue #416, opened 2019-11-21
Error in mapping config to dataset


jacobthill:
There seems to be a problem mapping some config files to data sets during transform. Running ```
docker run --rm -e SKIP_FETCH_DATA=true                 -v $(pwd)/.:/opt/traject                 -v $(pwd)/../dlme-metadata:/opt/traject/data                 -v $(pwd)/output:/opt/traject/output                 suldlss/dlme-transform:latest                 auc/p15795coll11
```
does grabs coll1 instead of coll11
```
```
"trajects": [
      "auc_coll11_config.rb"
    ],
    "paths": [
      "auc/p15795coll11"
    ],
```
This implicates many AUC collections which were probably getting transformed with the wrong config files. Since the transformations were mostly the same this wasn't apparent. Now that the configs are changing it became obvious.

For example, running:
```
docker run --rm -e SKIP_FETCH_DATA=true                 -v $(pwd)/.:/opt/traject                 -v $(pwd)/../dlme-metadata:/opt/traject/data                 -v $(pwd)/output:/opt/traject/output                 suldlss/dlme-transform:latest                 auc/p15795coll_12
```
should use the config file named `auc_coll11_config.rb` as stated in the metadata_mappings file:
```
{
    "trajects": [
      "auc_coll11_config.rb"
    ],
    "paths": [
      "auc/p15795coll_12"
```
Instead it seems to treat `auc_coll11_config.rb` and `auc_coll12_config.rb`as equivalents and since `auc_coll11_config.rb` is lower it is incorrectly used to transform all auc_coll1*_config.rb collections.

This is a blocker for continuing AUC mapping.

See https://github.com/sul-dlss/dlme-transform/pull/417 for testing.

Run:
```
docker run --rm -e SKIP_FETCH_DATA=true                 -v $(pwd)/.:/opt/traject                 -v $(pwd)/../dlme-metadata:/opt/traject/data                 -v $(pwd)/output:/opt/traject/output                 suldlss/dlme-transform:latest                 auc/p15795coll_12
```
and check the output. The transformations are coming from `auc/p15795coll_11`.

aaron-collier:
I suspect https://github.com/sul-dlss/dlme-transform/blob/db51c100219f9186ab981e548afe45be546a9c0b/lib/config_finder.rb#L36 is the issue. Investigating

aaron-collier:
[...]

PR #417, opened 2019-11-21
Revise AUC mapping

jacobthill:
## Why was this change made?

I encountered an [error](https://github.com/sul-dlss/dlme-transform/issues/416) in the process of mapping that needs to be resolved before I can continue


aaron-collier:
This is an interesting situation and possibly a bug in traject. We may want to schedule some pairing time around a possibly improved approach to this.

jacobthill:
Ok I'm available to pair any time this afternoon. I'm not sure how to determine whether the issue is caused by traject or our transform code.

jacobthill:
I'm inclined to ignore the code climate warning. This was an incredibly complex mapping challenge because the metadata was so inconsistent across collections. I don't see much value in adding a common config to put only the agg fields.

traject_configs/auc_first_date_for_range_config.rb
```diff
@@ -1,5 +0,0 @@
-# frozen_string_literal: true
-
-# AUC collection p15795coll8 has undesired date info except in first dc:date field
-to_field 'cho_date_range_norm', extract_oai('dc:date'), first_only, strip, auc_date_range
-to_field 'cho_date_range_hijri', extract_oai('dc:date'), first_only, strip, auc_date_range, hijri_range
```
ndushay: it's fine that you don't want this as a standalone config, but it was put here to solve a problem with coll8 being not like the others ... did that concern go away?

jacobthill:
Naomi, those lines were moved into the config for that collection including the comment

aaron-collier:
Approving / Merging, though I think this warrants a good amount of refactoring to clean up the duplication. There does seem to be some difference in fields, but also a lot of those differences are literals, which we may want to move into settings to limit code duplication.

PR #418, opened 2019-11-21
Do not add empty strings to language hashed values

### LINKED ISSUE #413: Raise error when no cho_title present
cho_title is a required field and an error should be raised if it is missing. Somehow 29 records in the cambridge collection passed transformation without raising an error. 

https://s3-us-west-2.amazonaws.com/dlme-metadata-dev/output-4c2913b0-f254-487d-b43b-fc866adb2e85.ndjson

aaron-collier:
## Why was this change made?

Fixes #413 

If the first value in the language hash is empty, raise an error.

N/A

### DIFF
--- a/lib/macros/each_record.rb
+++ b/lib/macros/each_record.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'byebug'
-
 module Macros
   # Macros for post-processing data
   module EachRecord
@@ -18,7 +16,6 @@ module Macros
           values.each do |value|
             case value
             when Hash
-              # byebug
               result[value[:language]] += value[:values].reject(&:nil?).reject(&:empty?)
             else
               result['none'] += Array(value)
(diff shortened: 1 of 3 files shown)


lib/macros/each_record.rb
```diff
@@ -16,7 +16,7 @@ def convert_to_language_hash(*fields)
           values.each do |value|
             case value
             when Hash
-              result[value[:language]] += value[:values]
+              result[value[:language]] += value[:values] unless value[:values]&.first&.empty?
```
mjgiarlo: `result[value[:language]] += value[:values].reject(&:empty?)`

lib/macros/each_record.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
ndushay: did you intend to leave this in here?

aaron-collier: DOH

PR #424, opened 2019-11-25
Update contract rule to raise error on NOT FOUND

### LINKED ISSUE #409: Fix NOT FOUND debugging tool
When mapping to a normalized value such as cho_language or cho_edm type if a value is not found in any translation map, the value 'NOT FOUND' is stored in the field and an error should be raised when that problematic record is transformed, breaking the transform process and returning the record data so the error can be fixed. This used to work and is now broken. We need to get it up and running again.

aaron-collier:
## Why was this change made?

Fixes #409

### DIFF
--- a/spec/lib/contracts/cho_spec.rb
+++ b/spec/lib/contracts/cho_spec.rb
@@ -126,6 +126,16 @@ RSpec.describe Contracts::CHO do
           end
         end
 
+        context 'when hash contains the value NOT FOUND' do
+          let(:cho) { { field_name => { 'none' => ['NOT FOUND'] } } }
+
+          it 'states the field included NOT FOUND' do
+            expect(contract.errors[field_name]).to include(
+              'unknown language found (NOT FOUND)'
+            )
+          end
+        end
+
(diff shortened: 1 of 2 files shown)


spec/lib/contracts/cho_spec.rb
```diff
@@ -126,6 +126,16 @@
           end
         end
 
+        context 'when hash contains the value NOT FOUND' do
+          let(:cho) { { field_name => { 'none' => ['NOT FOUND'] } } }
+
+          it 'states the field included NOT FOUND' do
```
ndushay: this is ok but a better description would be something like "errors include ..." however I see you were just following the pattern.

aaron-collier: Yeah, this likely could use a better description. I'll think about it. The purpose of this change is kind of the opposite of the other one. In the other one, we do not want empty values being added to the hashes because the validation doesn't catch it properly. In this one, we want to ensure that in the case that a bad language value is used (which is translated to NOT FOUND) the error is raised such that it can be fixed and not indexed.

PR #425, opened 2019-11-25
Add a debug writer to allow local debugging

aaron-collier:
## Why was this change made?

This adds a debug writer that can be used with the `-w` command line option. We currently fail in the out-of-the-box traject debug writer as our `serialize` method has been significantly customized.

This removes raising an error the DlmeJsonResourceWriter so that large collections do not fail on single records, but allows local development to utilize the `-w` option to fail on single records for debugging.

Bonus outcome is that this should fix the flappy integration test that is based on pulling `dlme-metadata`.

TBD: A small update to the README may be required.

### DIFF
--- a/spec/lib/dlme_debug_writer_spec.rb
+++ b/spec/lib/dlme_debug_writer_spec.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 require 'stringio'
-require 'dlme_json_resource_writer'
+require 'dlme_debug_writer'
 
-RSpec.describe DlmeJsonResourceWriter do
+RSpec.describe DlmeDebugWriter do
   let(:out) { StringIO.new }
   let(:settings) { { 'output_stream' => out } }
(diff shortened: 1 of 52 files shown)


spec/lib/dlme_debug_writer_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe '#put' do
+    subject(:put) do
+      writer.put(context)
+    end
+
+    context 'with invalid data' do
+      let(:data) do
+        { 'id' => ['one'], 'two' => %w[two1 two2], 'three' => 'three', 'four' => 'four' }
+      end
+      it 'logs an error' do
```
ndushay: whoops!  copy pasta in spec description!

aaron-collier: woops, fixed.

ndushay:
:-( build still failing 😢

PR #428, opened 2019-11-26
Add collection value as agg_data_provider_collection to IR

aaron-collier:
## Why was this change made?

Extract the collection name from the input file path. All file paths start with the `data_path` (usually `./data/`) so that is dropped as well as dropping the individual input file name as we want the path as a collection, not the individual file as a collection.

This adds:
- agg_data_provider_collection to the contract
- a macro to extract the collection name as the path
- a test for the macro
- appropriate config and `to_field` to each traject config

Note, this is perhaps a intermediate way to set a usable collection value in order to extrapolate the number of collections per provider in the spotlight stats. A future update may be necessary to implement a lookup to a displayable collection value, but this is the MVP for providing the statistic value in the app.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -7,6 +7,7 @@ fields:
       - agg_preview
       - transform_version
       - transform_timestamp
+      - agg_data_provider_collection
     web_resources:
       - agg_is_shown_at
       - agg_is_shown_by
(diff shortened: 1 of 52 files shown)


config/settings.yml
```diff
@@ -7,6 +7,7 @@ fields:
       - agg_preview
       - transform_version
       - transform_timestamp
+      - agg_data_provider_collection
```
ndushay: i think this is alphabetical

lib/contracts/cho.rb
```diff
@@ -41,6 +41,7 @@ class CHO < Dry::Validation::Contract
       required(:id).filled(:string)
       optional('__source'.to_sym).filled(:string)
       # Since the IR is a flattened projection of the MAP, 'agg_aggregated_cho' is not used.
+      required(:agg_data_provider_collection).value(:string)
```
ndushay: i think this is alphabetical as well.

PR #431, opened 2019-11-27
Draft Princeton reharfest, remap, add iiif

jacobthill:
## Why was this change made?

- Reharvest and remap Princeton data
- metadata changes on `princeton` branch of dlme-metadata

## Problems

- iiif integration needs checking/revision
- 'agg_preview' (line 126) is causing validation errors
- 'agg_is_shown_by' (line 112) might cause errors as well
- 'each_record_do' (line 29) is really slow since its harvesting the iiif manifest upon transfrom. Is there a better way to do this? Should these be stored with the metadata?
- service object methods (line 33...) are duplicated from `stanford_mods_config.rb`. Perhaps they should be broken out into a macro?
- need to add new language fields somewhere (check with Chris) so the new fields are indexed correctly


lib/macros/princeton.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting Stanford Specific MODS values from Nokogiri documents
+  module Michigan
```
aaron-collier: I'm assuming this should NOT be "Michigan"

traject_configs/princeton_mss_config.rb
```diff
@@ -22,58 +26,117 @@
...
   )
 end
-to_field 'agg_preview' do |_record, accumulator, context|
-  accumulator << transform_values(
-    context,
-    'wr_id' => [extract_json('.thumbnail'), strip]
-  )
-end
-
+# to_field 'agg_is_shown_by' do |_record, accumulator, context|
```
aaron-collier: Before moving this out of DRAFT, please remove this large comment.

aaron-collier:
@jacobthill I think this is likely ready (or very close to ready) to move out of draft. let's pair on questions in the morning and a few related topics to your list of problems above.

aaron-collier:
Closing in favor of #446

PR #446, opened 2019-12-09
Cleanup princeton mss PR

aaron-collier:
## Why was this change made?

Replaces #431 

Follow up PRs will be submitted for movie_posters if needed.

@jacobthill this is mostly a copy from your existing PR with a few tweaks and removing the change in the error message that I did (which broke the test). That should get fixed in a separate PR. I would suggest running this against your new princeton data and then merging that into dlme-metadata/master at the same time this is approved/merged. Also note, that I believe the build error here is specifically related to the dlme-metadata repo mismatch, so merging there first, then verifying the build works here before merging is probably the best approach.

N/A


jacobthill:
@aaron-collier I simplified the iiif portion of the princeton config. Please take a look and merge when you get a chance and then I can update the movie posters on a new PR.

lib/macros/iiif.rb
```diff
@@ -87,6 +87,12 @@ def iiif_service_conforms_to(service_profile)
         'http://iiif.io/api/presentation/'
       elsif service_profile.include? 'http://iiif.io/api/search/'
         'http://iiif.io/api/search/'
+      # this doesn't look right, but it's what we get.
+      # can be removed after: https://github.com/pulibrary/figgy/pull/3513
+      elsif service_profile.include? 'http;//iiiif.io/api/image/2/level2.json'
```
mjgiarlo: Looks like the figgy PR has already been merged. Can this be removed now?

lib/translation_maps/languages.yaml
```diff
@@ -182,6 +182,7 @@ thai: Thai
 tigrinya: Tigrinya
 tibetan: Tibetan
 turkish, ottoman: Turkish, Ottoman (1500-1928)
+urkish, ottoman (1500-1928): Turkish, Ottoman (1500-1928)
```
mjgiarlo: Is `urkish` a typo for `turkish`?

lib/macros/iiif.rb
aaron-collier: I'll verify, I did not initially see the change applied, so was not sure it was deployed or perhaps cached. But will check.

lib/translation_maps/languages.yaml
aaron-collier: Looks like. Will fix.

lib/macros/iiif.rb
aaron-collier: Confirmed and removed.

PR #451, opened 2020-01-06
Test Turkish subtypes

jacobthill:
## Why was this change made?
There are many Turkish values in the type field that need to be mapped to subtypes (cho_has_type). It is unclear what some of the values mean so I mapped some and passed others through to view them in the application and determine if a new sub type label should be added. 

n/a


jacobthill:
@aaron-collier I just need to get these loaded into the staging environment so I can figure out what some of the Turkish type values mean.

PR #461, opened 2020-01-23
Problem with setting default for missing title

jacobthill:


jacobthill:
The `default` transformation doesn't work anymore. If I don't add the `lang` transformation an error is not raised but a default value isn't set as it should be. If I do add the `lang` transformation I get a data validation error for missing title. Here is a record that should help reproduce the error:

```docker run --rm -e SKIP_FETCH_DATA=true                 -v $(pwd)/.:/opt/traject                 -v $(pwd)/../dlme-metadata:/opt/traject/data                 -v $(pwd)/output:/opt/traject/output                 suldlss/dlme-transform:latest                 cambridge/genizah/genizah-8042.xml```

PR #469, opened 2020-01-29
Fix language values

jacobthill:
## Why was this change made?

- Update language values for Shahre Farang records

- n/a


aaron-collier:
Not sure why we're seeing that oddly small total-coverage drop. Merging.

PR #518, opened 2020-03-04
update iiif manifests in LOC mapping

jacobthill:
## Why was this change made?

- Some iiif manifests didn't have canvases and needed to be removed
- Better approach for grabbing iiif manifests with canvases

n/a

### DIFF
--- a/traject_configs/loc_greek_and_armenian_patriarchates_config.rb
+++ b/traject_configs/loc_greek_and_armenian_patriarchates_config.rb
@@ -65,7 +65,7 @@ to_field 'agg_is_shown_at' do |_record, accumulator, context|
   accumulator << transform_values(
     context,
     'wr_id' => [extract_json('.id'), strip],
-    'wr_is_referenced_by' => [extract_json('.url'), strip, append('/manifest.json')]
+    'wr_is_referenced_by' => [extract_json('.id'), strip, append('/manifest.json')]
   )
 end
 to_field 'agg_preview' do |_record, accumulator, context|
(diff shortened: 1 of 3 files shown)


traject_configs/loc_greek_and_armenian_patriarchates_config.rb
```diff
@@ -65,7 +65,7 @@
   accumulator << transform_values(
     context,
     'wr_id' => [extract_json('.id'), strip],
-    'wr_is_referenced_by' => [extract_json('.url'), strip, append('/manifest.json')]
+    'wr_is_referenced_by' => [extract_json('.id'), strip, append('/manifest.json')]
```
jcoyne: This change seems to go beyond the description of the commit "remove iiif manifests that don't have canvases", Should more description be provided?

traject_configs/loc_persian_config.rb
```diff
@@ -67,8 +67,8 @@
 to_field 'agg_is_shown_at' do |_record, accumulator, context|
   accumulator << transform_values(
     context,
-    'wr_id' => [extract_json('.segments[0].url'), strip],
-    'wr_is_referenced_by' => [extract_json('.number_lccn[0]'), strip, prepend('https://www.loc.gov/item/'), append('/manifest.json')]
+    'wr_id' => [extract_json('.id'), strip],
+    'wr_is_referenced_by' => [extract_json('id'), strip, append('/manifest.json')]
```
jcoyne: This change seems to go beyond the description of the commit "remove iiif manifests that don't have canvases", Should more description be provided?

Issue #517, opened 2020-03-04
Dump all records from dlme prod. 


jacobthill:
I need to dump everything from DLME prod. The IR has changed and the records need to be re-transformed and re-loaded.

aaron-collier:
Done.

PR #521, opened 2020-03-05
Fix broken agg_preview, agg_is_shown_at urls and bad iiif manifests

jacobthill:
#517  Why was this change made?
- fix bad urls for agg_preview
- fix bad urls for agg_is_shown_at
- remove bad iiif manifests for both

n/a


jacobthill:
@peetucket that is the problem with the current tranform pipeline. We don't have tests for everything. I have written my own analysis code in Python and am fairly certain this will fix some problems. There may be others but it won't be clear until I load it in dev and examine it.

peetucket:
ok, go ahead then..

PR #531, opened 2020-03-30
Princeton

### LINKED ISSUE #532: Detect script/characeter set from text of title and use it to call another macro to add language info to the title field
We need to examine the characters that exist in the title of an incoming record and use this to pass to a method to set the script-language pair.

e.g. a macro that takes a record as input and returns the script type as an output as shown here: https://github.com/sul-dlss/dlme-transform/pull/531

and then call the existing `lang()` macro with the result, like here but dynamically: https://github.com/sul-dlss/dlme-transform/pull/531/files#diff-b1586f1b2ce8cc5cd2e78648ee99c792R37

so that `cho_title` has the language info

peetucket:
## Why was this change made?

fixes #532

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -32,7 +32,8 @@ acceptable_bcp47_codes:
   - ur-Latn
 defaults:
   base_data_dir: 'data'
-#  base_data_dir: '../dlme-metadata' # uncomment if needed when running tests locally
+#  base_data_dir: '../dlme-metadata' # if you haven't cloned the dlme-metadata repo into the `data` subfolder
+#                                      you can uncomment this line if needed when running tests locally (but don't commit)
   data_dir: ''
   mapping_file: 'config/metadata_mapping.json'
   traject_dir: 'traject_configs'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -24,7 +24,8 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
-  - ms
+  - ms-Arab
+  - ms-Latn
   - tr-Arab
   - tr-Latn
   - ur-Arab
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -32,6 +32,7 @@ acceptable_bcp47_codes:
   - ur-Latn
 defaults:
   base_data_dir: 'data'
+#  base_data_dir: '../dlme-metadata' # uncomment if needed when running tests locally
   data_dir: ''
   mapping_file: 'config/metadata_mapping.json'
   traject_dir: 'traject_configs'
--- a/lib/macros/princeton.rb
+++ b/lib/macros/princeton.rb
@@ -11,7 +11,7 @@ module Macros
     end
 
     def get_title_language(record)
-      lang_map = { ara: 'ar' }
+      lang_map = { ara: 'ar' } # other values needed here to map from incoming data to output expected?
       lang_map[record.dig('title').first.dig('@language').to_sym]
     end
   end
--- a/lib/macros/princeton.rb
+++ b/lib/macros/princeton.rb
@@ -3,14 +3,9 @@
 module Macros
   # Macros for extracting language and derecting script from title
   module Princeton
-
-    def get_language_rec
+    def language_rec
       lambda do |record, accumulator|
-        if accumulator.first.match?(/[aeiou]/)
-          script = 'Latn'
-        else
-          script = 'Arab'
-        end
+        script = accumulator.first.match?(/[aeiou]/) ? 'Latn' : 'Arab'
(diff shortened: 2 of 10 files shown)


lib/macros/princeton.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting Stanford Specific MODS values from Nokogiri documents
```
ndushay: comment needs love - not stanford, and not MODS and not nokogiri?

peetucket: thanks, adjusted

peetucket:
I was having trouble running the tests to debug the failure (even cloned the entire dlme-metadata but the tests took forever to run and seemed to be spitting out way too much transformed data to output...maybe I don't know how to run.

I think the failure is perhaps a result of needing to a nil check in the macro and return some default language-script pair.

lib/macros/princeton.rb
```diff
@@ -0,0 +1,39 @@
...
+  # Macros for extracting language and derecting script from title
+  module Princeton
+    def princeton_title_and_lang
+      lambda do |record, accumulator|
+        begin
+          title = record.dig('title').first.dig('@value')
+          script = title.match?(/[aeiou]/) ? 'Latn' : 'Arab'
+          language = get_title_language(record)
+          language_and_script = "#{language}-#{script}"
+        rescue
```
peetucket, line 13: multiple rescue blocks is not a great idea ... a better approach is to do some conditional checks (like a series of if..else clauses) on the title value in the record instead

if we really need a single rescue clause at the end as a catchall, we should also only rescue the specific exception that may occur and not everything

peetucket: i made some local changes to this effect, happy to share (but don't want to immediately push my changes in case they conflict with work you may be continuing on)

spec/integration/traject_configs_spec.rb
```diff
@@ -19,7 +19,7 @@
           settings: config.fetch('settings', {})
         ).transform
       rescue StandardError => e
-        raise "error raised mapping #{config['settings']['inst_id']}: #{e.class}: #{e.message}"
+        raise "error raised mapping #{config['settings']['inst_id']} for #{data_filepath}: #{e.class}: #{e.message}"
```
peetucket, line 22: note: this is helpful when running the tests as it shows the file that caused a failure

config/settings.yml
```diff
@@ -24,13 +24,15 @@ acceptable_bcp47_codes:
...
-  - ms
+  - ms-Arab
+  - ms-Latn
   - tr-Arab
   - tr-Latn
   - ur-Arab
   - ur-Latn
 defaults:
   base_data_dir: 'data'
+#  base_data_dir: '../dlme-metadata' # uncomment if needed when running tests locally
```
peetucket, line 35: note: i found this helpful when running the tests locally against a local cloned dlme-metadata directory

jacobthill: This is helpful. I think I came up with the same solution [here](https://github.com/sul-dlss/dlme-transform/blob/master/lib/macros/path_to_file.rb) if I understand this correctly but this might be a better way of providing the file information. Do you see any reason to not leave this on? I can't think of any problems it would cause and it would be unfortunate to transform 20k records and then realize you should have left this setting on and have to run it again.

peetucket: This isn't the change to show the file that failed the test - but I needed to change this temporarily to run my tests locally (since all of the files were in a dlme-metadata directory one level up instead of the 'data' subfolder).  I suppose I could just clone the dlme-metadata directory in the data subfolder instead (since the entire contents of the data subfolder are ignored).  I am going to try that.

peetucket: This is the change that shows the file when running the tests: https://github.com/sul-dlss/dlme-transform/pull/531/files#diff-d37cfe808b73c055a3280096ffa98e5cR22

jacobthill:
@peetucket can we pair on writing the tests for the naive_language_extractor macro I added? I'm not sure about how to fix these circleci errors.

peetucket:
> @peetucket can we pair on writing the tests for the naive_language_extractor macro I added? I'm not sure about how to fix these circleci errors.

sure, let's connect in slack and find a time that works for both of us ... i took a quick look at those failing tests and its not clear what it is that is being tested to me, but `instance` in this case is the class itself, so I think you need to call the actual method you are wanting to test to confirm its output?  similar to what is happening above on lines 35-38

it is a bit confusing since the `lang` method returns a Proc, which then needs to be called (that is what is happening in the test on line 37-38)

`expect(instance).to eq([{ language: 'ar-Arab', values: [extracted_string] }])`

peetucket:
I think I got a lead, let's pair later and I can show you what I did.

jacobthill:
Thanks @peetucket. Just ping me on slack when you're ready.

PR #554, opened 2020-05-27
Update numismatic_csv_config.rb

ewg118:
Updated geography, creator, contributor. Removed era (too inconsistent). Removed manufacture, weight (not useful for a query of this sort of aggregated content).


ewg118:
Was the IIIF manifest URL removed because they are http and not https?

jacobthill:
Thanks @ewg118 for this. You are the first DLME data partner to submit a PR. The current data we are using for American Numismatic Society is [here](https://github.com/sul-dlss/dlme-metadata/tree/master/american-numismatic-society/islamic-department/data). The data we have doesn't seem to have any values for `Engraver` or `Issuer` and it doesn't have the `Locality` field. Do you have more up-to-date data? If so you can submit a PR in https://github.com/sul-dlss/dlme-metadata.

jacobthill:
Also I can't recall anything about the iiif manifest. Maybe that was before my time. Our data doesn't have a iiif manifest. If you can provide the urls to the manifests, I can explore that.

ewg118:
Our database is incredibly inconsistent from department to department, so it doesn't look like Locality, Engraver, or Issuer have been used. It appears that about 700 records have been updated since October, so some data or photos might be added or updated. Assuming that the original dump consisted of our Islamic department including only photographed coins (http://numismatics.org/search/results?q=department_facet%3A%22Islamic%22+AND+imagesavailable%3Atrue), it's possible to download CSV by clicking the icon on the top-left corner (data options). It is limited to 1000 per response, but you can paginate by appending `&start=1000` for the next batch.

ewg118:
Updated the mapping.

The manifest URL is not stored in our CSV export, but it follows a pattern of `http://numismatics.org/search/manifest/{recordId}`, e.g., http://numismatics.org/search/manifest/1944.100.52684

jacobthill:
Closing in favor of https://github.com/sul-dlss/dlme-transform/pull/559. For some reason the tests are stuck.

PR #568, opened 2020-06-15
revise harvard scw

jacobthill:
## Why was this change made?

The dates displayed were redundant and the rights statement applied to the physical object not the digital images. There is no statement for the digital images.

tested locally

n/a

### DIFF
--- a/traject_configs/harvard_scw_config.rb
+++ b/traject_configs/harvard_scw_config.rb
@@ -47,21 +47,9 @@ to_field 'id', generate_mods_id
 to_field 'cho_alternative', extract_mods('/*/mods:titleInfo[@type]/mods:title'), lang('en')
 to_field 'cho_coverage', extract_mods('/*/mods:originInfo/mods:place/mods:placeTerm'), lang('en')
 to_field 'cho_creator', extract_name(role: %w[author creator]), lang('en')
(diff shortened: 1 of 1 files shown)


traject_configs/harvard_scw_config.rb
```diff
@@ -47,21 +47,9 @@
 to_field 'cho_alternative', extract_mods('/*/mods:titleInfo[@type]/mods:title'), lang('en')
 to_field 'cho_coverage', extract_mods('/*/mods:originInfo/mods:place/mods:placeTerm'), lang('en')
 to_field 'cho_creator', extract_name(role: %w[author creator]), lang('en')
-to_field 'cho_date', extract_mods('/*/mods:originInfo/mods:dateCreated')
+to_field 'cho_date', extract_mods('/*/mods:originInfo/mods:dateCreated[3]')
```
aaron-collier, line 50: Will this always be the 4th dateCreated field?

jacobthill: Indexing starts with 1 in this case but yes it will always be the third instance according to the data provider. I checked by manually inspecting a few dozen records and it was consistent in those cases. They always use the first for their range start year and the second for their range end year. The third is meant for human consumption and adds in things like c. to indicate certainty, etc.

PR #588, opened 2020-07-29
Map LOC collections

jacobthill:
## Why was this change made?

These collections are new to DLME. Mapping started a few months back but wasn't finished. Maps six collections, adds new language detection macros to assign language values to metadata fields (these will be broken out of dlme.rb in the future), updates metadata settings (one country intentionally left blank) and translation map.

Locally tested as much as possible in the docker app. Also used mapping analysis scripts.

n/a

### DIFF
--- a/traject_configs/loc_persian_config.rb
+++ b/traject_configs/loc_persian_config.rb
@@ -41,7 +41,6 @@ to_field 'cho_title', extract_json('.item.title'), strip, persian_or_none
 to_field 'cho_title', extract_json('.item.other_title[0]'), strip, persian_or_none
 
 # Cho Other
-# to_field 'cho_alternative', extract_json('.item.other_title[0]'), strip
 to_field 'cho_contributor', extract_json('item.contributors[0]'), strip, lang('en')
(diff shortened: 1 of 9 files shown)


traject_configs/loc_persian_config.rb
```diff
@@ -30,12 +32,16 @@
...
+
 # Cho Required
 to_field 'id', extract_json('.id')
-to_field 'cho_title', extract_json('.item.title'), strip
+to_field 'cho_title', extract_json('.item.title'), strip, persian_or_none
+to_field 'cho_title', extract_json('.item.other_title[0]'), strip, persian_or_none
 
 # Cho Other
-to_field 'cho_alternative', extract_json('.item.other_title[0]'), strip
+# to_field 'cho_alternative', extract_json('.item.other_title[0]'), strip
```
aaron-collier, line 44: Can this line be removed or is there a reason to keep the comment?

jacobthill: Good catch, I just removed it.

aaron-collier: LGTM, pending green.

PR #596, opened 2020-08-19
Update README.md

jacobthill:
Change data path (the old path no longer exists) and remove extra 'arn:'

## Why was this change made?

There were a couple errors that I didn't catch before merging.

locally

README


aaron-collier:
Good catch!

PR #598, opened 2020-08-26
Map Openn collections

jacobthill:
## Why was this change made?

The OPenn project has added several more collections and changes the data model as well. These were re-harvested and the mappings needed to be revised.

locally through analysis scripts and in the dev environment 

n/a


jacobthill:
@aaron-collier if you have a moment, can you let me know what you think we should do about these warnings? The cognitive complexity issue is the same in `cambridge_gregorian_range` which we ignored and pushed through. I think the intention was to review it at another time. The failed `el-taher` config is expected. That was merged a few weeks ago to test the data. I need to return review and revise the loc configs.

PR #602, opened 2020-09-22
fix edm_type

jacobthill:
## Why was this change made?

Fix a few errors in LOC el-Taher collection
- remove trailing backslash in title field
- one photograph had a cho_edm_type value of 'Text' instead of 'Image'.

Local transform viewed in the dev site.

loc_el_taher_config.rb, no documentation update necessary

### DIFF
--- a/traject_configs/loc_el_taher_config.rb
+++ b/traject_configs/loc_el_taher_config.rb
@@ -39,8 +39,8 @@ to_field 'dlme_source_file', path_to_file
 
 # Cho Required
 to_field 'id', extract_json('.id')
-to_field 'cho_title', extract_json('.item.title'), strip, arabic_or_none
-to_field 'cho_title', extract_json('.item.other_title[0]'), strip, arabic_or_none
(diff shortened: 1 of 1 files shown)


traject_configs/loc_el_taher_config.rb
```diff
@@ -39,8 +39,8 @@
 
 # Cho Required
 to_field 'id', extract_json('.id')
-to_field 'cho_title', extract_json('.item.title'), strip, arabic_or_none
-to_field 'cho_title', extract_json('.item.other_title[0]'), strip, arabic_or_none
+to_field 'cho_title', extract_json('.item.title'), strip, gsub('/', ''), arabic_or_none
```
jcoyne, line 42: Should this be in a separate PR?  The PR says this is about edm_type, but this is cho_title?

jcoyne (changes requested):

traject_configs/loc_el_taher_config.rb
jacobthill: I think its fine in one PR but I have changed the description to reflect the correct scope.

jcoyne (approved):

PR #622, opened 2020-11-10
Add Lausanne mapping

jacobthill:
## Why was this change made?

Adding a new collection 

Locally and viewed in dev

n/a

### DIFF
--- a/lib/macros/lausanne.rb
+++ b/lib/macros/lausanne.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting values from Lausanne records
+  module Lausanne
+    # Extracts earliest & latest dates from Lausanne record and merges into singe date range value
+    def lausanne_date_range
+      lambda do |record, accumulator, context|
+        first_year = record['from'].to_i if record['from']&.match(/\d+/)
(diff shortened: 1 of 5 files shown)


lib/macros/lausanne.rb
```diff
@@ -0,0 +1,24 @@
...
+        accumulator.replace(range_array(context, first_year, last_year))
+      end
+    end
+
+    # Extracts earliest & latest dates from Lausanne record and merges into singe date string value
+    def lausanne_date_string
+      lambda do |record, accumulator|
+        first_year = record['from'] if record['from']&.match(/\d+/)
+        last_year = record['to'] if record['to']&.match(/\d+/)
+        accumulator.replace(["#{first_year} - #{last_year}"]) if record['from'] || record['to']
```
aaron-collier, line 20: In this case, if either of the values is `nil` you will end up with ` - last_year` or `first_year -`, is that OK?

jacobthill: Yeah, its not ideal. I guess I should switch that to `&&`

aaron-collier: if ONLY the `first_year - last_year` is valid, we could probably simplify this so you don't need the last if... i'll do a PR against this branch with a suggestion.

aaron-collier (approved):

PR #625, opened 2020-11-12
Simplify the lausanne macros to limit complexity

aaron-collier:
## Why was this change made?

A suggestion to fix the complexity violation of the lausanne macros.


jacobthill (approved):

jacobthill (changes requested):
Seems like it broke the tests.

PR #627, opened 2020-11-16
Refactor language extraction macros, fixes #617

### LINKED ISSUE #617: Refactor language extraction macros
Rather than have a macro for every combination of languages (e.g. arabic_or_english, arabic_or_none, etc.) have arabic_or_default, english_or_default, etc. where the default language can be passed as parameter.
```
    def arabic_or_default(default = 'none')
        accumulator.map! do |val|
          script = val.match?(/[ضصثقفغعهخحمنتالبيسشظطذدزرو]/) ? 'ar-Arab' : default
          { language: script.to_s, values: [val] }
        end
      end
    end
```

jacobthill:
## Why was this change made?

Multiple language extraction macros with hard coded lang values needed to be refactored into one macro where the langs are passed as arguments.

local transform

n/a


traject_configs/loc_el_taher_config.rb
```diff
@@ -39,8 +39,8 @@
 
 # Cho Required
 to_field 'id', extract_json('.id')
-to_field 'cho_title', extract_json('.item.title'), strip, gsub('/', ''), arabic_or_none
-to_field 'cho_title', extract_json('.item.other_title[0]'), strip, gsub('/', ''), arabic_or_none
+to_field 'cho_title', extract_json('.item.title'), strip, gsub('/', ''), arabic_script_lang_or_default('ar-Arab', 'und-Latn')
+to_field 'cho_title', extract_json('.item.other_title[0]'), strip, gsub('/', ''), arabic_script_lang_or_default('ar-Arab', 'und-Latn')
```
jmartin-sul, line 43: based solely on the code being refactored, it seems like the second arg value for the new call here should be `'none'`?  or maybe i'm mis-reading, or maybe this is an intentional behavioral change? (not super familiar w/ this codebase, so this could be fine and it could be me)

traject_configs/loc_persian_config.rb
```diff
@@ -39,8 +39,8 @@
 
 # Cho Required
 to_field 'id', extract_json('.id')
-to_field 'cho_title', extract_json('.item.title'), strip, persian_or_none
-to_field 'cho_title', extract_json('.item.other_title[0]'), strip, persian_or_none
+to_field 'cho_title', extract_json('.item.title'), strip, arabic_script_lang_or_default('fa-Arab', 'und-Latn')
+to_field 'cho_title', extract_json('.item.other_title[0]'), strip, arabic_script_lang_or_default('fa-Arab', 'und-Latn')
```
jmartin-sul, line 43: similar question here, should second param be `'none'`?

traject_configs/openn_mmw_0043_config.rb
```diff
@@ -58,15 +58,15 @@
   bare_id = default_identifier(context)
   accumulator << identifier_with_prefix(context, bare_id)
 }
-to_field 'cho_title', extract_tei("#{MS_DESC}/#{MS_CONTENTS}/#{MS_ITEM}[1]/tei:title"), split('/'), last, strip, arabic_or_und_latn
+to_field 'cho_title', extract_tei("#{MS_DESC}/#{MS_CONTENTS}/#{MS_ITEM}[1]/tei:title"), split('/'), last, strip, arabic_script_lang_or_default('ar-Arab', 'en')
```
jmartin-sul, line 61: should param value 2 be `'und-Latn'`?

jmartin-sul (commented):
questions about some of the param values being fed to some of the refactored method calls, might be an intentional change or just me mis-understanding things due to lack of familiarity w/ this project.  otherwise, LGTM.

traject_configs/openn_mmw_0043_config.rb
jacobthill: This macro provides a simple way for assigning lang tags based on script. It checks for Arabic character, if present assigns the first arg, else assigns the second arg. The args depend on the data set. In this case the title values are either in Arabic language in Arabic script ('ar-Arab') or they are in English ('en').

jmartin-sul (approved):
discussed on slack, there were some intended behavioral changes here, all good.

PR #628, opened 2020-11-19
Cambridge

jacobthill:
## Why was this change made?

- remap Cambridge records based on new input from QNL
- add new string helper macros
- break string helper macros out of DLME module into new StringHelper module because DLME was too long
- add `cambridge_dimensions` macro to extract dimensions field into formatted string per QNL request
- begin `agg_edm_rights` mapping, add translation map to support this
- add `return_or_prepend` macro to safely prepend without prepending to an empty array
- add new lang code and update dlme https://github.com/sul-dlss/dlme/pull/1089

- local transform and load in dev

- n/a


config/settings.yml
```diff
@@ -24,6 +24,7 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
+  - he
```
jmartin-sul, line 27: comment on the section mentions "if these values are changed, they need to be synced with configuration in DLME", but didn't see this value there or a PR to put it in.

lib/macros/string_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+  module StringHelper
+    # Returns a string with new lines and whitespace removed.
+    # @return [Proc] a proc that traject can call for each record
+    # @example
+    # "Part of the \n      Islamic Manuscripts |n    Collection" => "Part of the Islamic Manuscripts Collection"
+    def squish
+      lambda do |_rec, acc|
+        acc.collect! do |v|
+          # unicode whitespace class aware
+          v.gsub(/\s+/, ' ')
```
jmartin-sul, line 14: fwiw, `activesupport` is already installed in this project, and there's an `activesupport` method (added to `String` objects) for this:  https://apidock.com/rails/v5.2.3/String/squish (though if you view the source, it's doing pretty much the same thing you are here, and i don't have a strong preference for switching to the `activesupport` call)

lib/macros/string_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    # Returns a string with first letter of each word capitalized.
+    # @return [Proc] a proc that traject can call for each record
+    # @example
+    # "the qur'an" => "The Qur'an"
+    def titleize
+      lambda do |_rec, acc|
+        acc.collect! do |v|
+          v.split(/(\W)/).map(&:capitalize).join
```
jmartin-sul, line 26: ditto this https://apidock.com/rails/v5.2.3/String/titleize

lib/macros/string_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+      end
+    end
+
+    # Shorten a string and follow it with an ellipsis.
+    def truncate(text, length = 100, truncate_string = '...')
```
jmartin-sul, line 32: and https://apidock.com/rails/v5.2.3/String/truncate

spec/lib/traject/macros/dlme_spec.rb
```diff
@@ -65,32 +67,36 @@
...
     end
 
-    context 'when extracted string shorter than 100 charachters' do
-      it 'returns string without change' do
-        arabic_string = 'تتعلق المراسلات وأوراق أخرى'
-        latin_string = 'Euchologion ad usum Melchitarum'
-        expect(instance.truncate(arabic_string)).to eq('تتعلق المراسلات وأوراق أخرى')
-        expect(instance.truncate(latin_string)).to eq('Euchologion ad usum Melchitarum')
+    context 'when value missing' do
+      let(:title) { '<title><title>' }
```
jmartin-sul, line 88: should this be `'<title></title>'`?

spec/lib/traject/macros/dlme_spec.rb
```diff
@@ -65,32 +67,36 @@
...
+
+      it 'has title provided a value' do
+        expect(indexer.map_record(ng_rec)).to eq({})
+      end
+    end
+
+    context 'when value present' do
+      let(:title) { '<title>Title<title>' }
```
jmartin-sul, line 96: should this be `'<title>Title</title>'`?

lib/macros/dlme.rb
```diff
@@ -147,10 +147,28 @@ def provider_country_ar
...
+    # Extract a OAI Dublin Core title or, if no title in record, extract abridged description, else pass default values.
+    def xpath_common_title_or_desc(xpath_title, xpath_desc, xpath_id)
+      lambda do |rec, acc|
+        title = rec.xpath(xpath_title, NS).map(&:text).first
+        description = rec.xpath(xpath_desc, NS).map(&:text).first
+        id = rec.xpath(xpath_id, NS).map(&:text).first
+        if title.present?
+          acc.replace(["#{title} #{id}"])
+        elsif description.present?
+          acc.replace([truncate(description)])
```
jmartin-sul, line 169: a test for this code path might be nice (though i'm not sure i'd have noticed it was uncovered if the codeclimate plugin hadn't highlighted it)

jmartin-sul (commented):
some minor questions and nitpicks, but this mostly looks fine to me.  i'm not really comfortable reviewing the traject configs part, since i'm not very familiar with that, so probably best to get someone a bit more familiar with the project to sign off before merging.

spec/lib/traject/macros/dlme_spec.rb
jacobthill: yep

config/settings.yml
jacobthill: I will put in a PR for this.

lib/macros/string_helper.rb
jacobthill: I was aware of all of these methods in `activesupport` but I need a hint on how to use them. I'm not sure how to make the methods available in a config. `extend ActiveSupport` isn't enough. I also might need to figure out how to call them in the traject pipeline.

lib/macros/dlme.rb
jacobthill: I'll wait on this to see if I switch to using the equivalents from `activesupport`

lib/macros/string_helper.rb
jacobthill: fixed

lib/macros/string_helper.rb
jacobthill: fixed

lib/macros/string_helper.rb
jacobthill: The active support truncate seems to work a little differently. Passing it in like we do with titleize and squish doesn't work in the traject pipeline so keeping this macro as is for now.

lib/macros/dlme.rb
jacobthill: I believe the test is there already. Am I missing something?

aaron-collier: Yeah - CodeClimate is showing that this branch is not covered by the existing test... Not all that big of a deal at this point really as it's so similar to the other path. I'm ok with a test coming in later.

lib/macros/cambridge.rb
```diff
@@ -1,11 +1,81 @@
 # frozen_string_literal: true
 
 module Macros
-  # Macros for extracting Stanford Specific MODS values from Nokogiri documents
+  # Macros for extracting Cambridge Specific values from Nokogiri documents
   module Cambridge
```
aaron-collier, line 5: Not suggesting a change in this PR, but this method would be a good example of doing a follow up refactoring PR as there is a lot of duplication here.

aaron-collier (approved):

PR #634, opened 2020-12-09
Update QNL mapping

jacobthill:
## Why was this change made?

Need to revise QNL mapping and add unused fields that are in the DLME application profile but have not yet been used.

Local transform

n/a


jacobthill:
This is a somewhat complex mapping. It adds several fields that I have not mapped before. I want to be sure that I am mapping the correct values. Can someone with a good understanding of iiif let me know if this example record is correct:
 
"agg_is_shown_at": 
  {"wr_dc_rights":["Open Government Licence"],
  "wr_edm_rights":["BY-SA"],
  "wr_format":["image/jpeg"],
  "wr_is_referenced_by":["https://www.qdl.qa/en/iiif/81055/vdc_100000000159.0x0003e3/manifest"],
  "wr_id":"https://www.qdl.qa/en/mirador/81055/vdc_100040046184.0x000061"},
"agg_is_shown_by":
  {"wr_edm_rights":["BY-SA"],
  "wr_format":["image/jpeg"],
  "wr_is_referenced_by":["https://www.qdl.qa/en/iiif/81055/vdc_100040046184.0x000061/manifest"],
  "wr_id":"https://iiif.qdl.qa/iiif/images/81055/vdc_100000000159.0x0003e3/IOR_R_15_1_15_0001.jp2"},
"agg_preview":
  {"wr_edm_rights":["BY-SA"],
  "wr_format":["image/jpeg"],
  "wr_is_referenced_by":["https://www.qdl.qa/en/iiif/81055/vdc_100040046184.0x000061/manifest"],
  "wr_id":"https://iiif.qdl.qa/iiif/images/81055/vdc_100000000159.0x0003e3/IOR_R_15_1_15_0001.jp2/full/80,100/0/default.jpg",
  "wr_has_service":[{"service_conforms_to":["http://iiif.io/api/image/2/level1.json"],
  "service_id":"https://iiif.qdl.qa/iiif/images/81055/vdc_100000000159.0x0003e3/IOR_R_15_1_15_0001.jp2",
  "service_implements":"http://iiif.io/api/image/2/level1.json"}

I'm asking for someone to look at the iiif manifest and confirm that for this record the values I mapped are correct.

lib/macros/iiif.rb
```diff
@@ -78,13 +87,16 @@ def rep_iiif_resource(manifest_json)
     end
 
     def iiif_service_conforms_to(service_profile)
+      puts service_profile
```
aaron-collier, line 90: Is this desired in the production code or was it debugging cruft?

lib/macros/qnl.rb
```diff
@@ -42,6 +45,24 @@ def extract_qnl_identifier
...
+        default_identifier(context)
+      end
+    end
+
+    # Grab the identifier from the MODS XML, or if one cannot be found, from the default.
+    # @param [Nokogiri::Document] record the MODS xml
+    # @param [Traject::Indexer::Context] context
+    # @return [String] the identifier
+
+    # def generate_druid(record, context)
```
jcoyne, line 61: Can this be removed?

lib/macros/qnl.rb
```diff
@@ -14,6 +14,9 @@ module QNL
     private_constant :AR_PREFIX
     private_constant :EN_PREFIX
 
+    # OAI_NS = 'http://www.openarchives.org/OAI/2.0/'
+    # private_constant :OAI_NA
```
jcoyne, line 18: Can this be removed?

lib/translation_maps/edm_rights.yaml
```diff
@@ -1,3 +1,4 @@
 # Basic mapping from DLME edm rights codes to value displayed in application.
-BY-NC-SA: 'CC BY-NC-SA: http://creativecommons.org/licenses/by-nc-sa/4.0/'
-InC-EDU: 'InC-EDU: http://rightsstatements.org/vocab/InC-EDU/1.0/'
+BY-NC-SA: 'CC BY-NC-SA'
+InC-EDU: 'InC-EDU'
+Open Government Licence: 'BY-SA'
```
jcoyne, line 4: I don't understand this mapping. Was this in error?

jcoyne (changes requested):

lib/translation_maps/edm_rights.yaml
jacobthill: Not an error. I believe these are the fields we want. The `InC-EDU: 'InC-EDU'` line is in because of the way we set up our test. I want to make sure all values are normalized and the only way to do that is to use the not-found macro that warns me if a value present in the metadata is not found in the translation map. I'm sure there is a better way to test this but this is what we have for now. That reminds me though that I need to call the not found macro. I will remove the cruft and add that line shortly.

jcoyne: I'm speaking specifically to line 4. The abbreviation for "Open Government License" is "OGL-UK-2.0" (per https://spdx.org/licenses/OGL-UK-2.0.html). "BY-SA" is "Creative Commons Attribution-ShareAlike"

lib/macros/qnl.rb
jacobthill: Removed.

lib/macros/qnl.rb
jacobthill: removed

lib/translation_maps/edm_rights.yaml
jacobthill: We are mapping to the [Europeana rights](https://pro.europeana.eu/page/available-rights-statements) for the purpose of downstream harvesting. Europeana doesn't have an Open Government License option so I mapped it to what seemed like the most appropriate functional equivalent. Essentially both options mean a user can re-use the content, even for commercial purposes.

jcoyne: I would suggest documenting that here so that people reading the code understand why this mapping is being made.

anarchivist: @jacobthill, how will these display in DLME? For what it's worth, the statements from the  Europeana rights statements set are usually referred to by their URI (e.g. `InC-EDU` should be `http://rightsstatements.org/vocab/InC-EDU/1.0/`). 

The Open Government License is distinct from, but compatible with CC-BY. If we do have things that are available under OGL, we should probably revise the validation rules and data mapping to ensure the proper license is displayed.

jacobthill: @anarchivist I'm not sure but I assumed they would not display in DLME. My assumption was that we would display wr_dc_rights but only use wr_edm_rights to determine what we could share with users via the API. At least that was my understanding of the conversation we had before break. Another option could be to deviate from the EDM model and add and Open Government License. What do you think? All of these different rights fields are a bit confusing to me.

anarchivist: @jacobthill OK, got it. In this case `wr_edm_rights` should be a URI according to the application profile. The set of Europeana rights statements that you linked to above isn't a direct constraint within the EDM model, so we should be OK here. Happy to go over this more at standup if you have additional questions.

lib/translation_maps/edm_rights.yaml
```diff
@@ -1,3 +1,8 @@
-# Basic mapping from DLME edm rights codes to value displayed in application.
+# Basic mapping from data provider values to Europeana rights codes to value displayed in application.
+# Codes not found in the Europeana model will be mapped to a functional equivalent. 
+# These codes will be used to determine downstream access.
+
 BY-NC-SA: 'CC BY-NC-SA: http://creativecommons.org/licenses/by-nc-sa/4.0/'
-InC-EDU: 'InC-EDU: http://rightsstatements.org/vocab/InC-EDU/1.0/'
+InC-EDU: 'InC-EDU: http://rightsstatements.org/page/InC-EDU/1.0/'
+Open Government Licence: 'BY-SA: https://creativecommons.org/licenses/by-sa/4.0/'
+Open Government License: 'BY-SA: https://creativecommons.org/licenses/by-sa/4.0/'
```
jcoyne, line 8: duplicate line here?

jacobthill: License is misspelled in some of the metadata.

jcoyne: Perhaps add a comment about that here too so that a reader will appreciate the intention.

jcoyne (approved):

PR #648, opened 2021-02-12
Harvard scw

jacobthill:
## Why was this change made?

- Breaks title extraction macros out of dlme Module to avoid too many lines warning, moves them to TitleExtraction module and updates configs.
- Adds `xpath_commas_with_prepend` macro to prepend to array instead of to each value of an array
- Adds new macros for Harvard IHP and SCW sets and breaks them into two modules to avoid too many lines warning
- Revise `extract_name` macro to take xpath as an argument; makes it useable with different xpath
- Adds new translation maps to catch unusual fields in these collections
- Updates tests
- These collections need to be reviewed by QNL and may need to be revised after. Better tests will be written once the mapping strategy is approved by QNL.

Local transforms and load in dev

Google docs partner profile


jacobthill:
The strategy used in the `ihp_has_type` macro needs to be approved by QNL. I would suggest ignoring the cognitive complexity issue. It the strategy is approved the macro can be revised, if needed. The `extract_ihp_manifest` didn't seem worthy of a test.

aaron-collier (approved):

PR #666, opened 2021-03-29
adjust title extraction

jacobthill:
## Why was this change made?

Fixes title values

Local transform and loaded in dev

n/a


jacobthill:
I am not sure about that circleCI error. Traject is not raising any errors locally and there doesn't appear to be a missing space before or after the comma. Maybe circleCI is having trouble interpreting the Arabic script with parenthesis?

ndushay:
Here is a way to turn off a particular rubocop rule so the lines won't trigger a rubocop exception, but I'm worried that if rubocop couldn't recognize where the comma was in relation to the Arabic strings ... will running the code have the same confusion?   Have you tested it?

```
# rubocop:disable Layout/SpaceBeforeComma
# rubocop:disable Layout/SpaceAfterComma
to_field 'cho_title', xpath_title_or_desc("#{PREFIX}/dc:title[last()]", "#{PREFIX}/dc:description[last()]"), gsub('﴾مجاهد ﴿شورا' ,'مجاهد (شورا)'), lang('fa-Latn')
# rubocop:enable Layout/SpaceBeforeComma
# rubocop:enable Layout/SpaceAfterComma
```

jacobthill:
Fixed the rubocop error. Should be good now.

peetucket (commented):
Putting aside jokes about this devil number PR, was it intended to include a new project config: `lausanne_tiresias_config.rb`?

jacobthill:
@peetucket that config wasn't supposed to be there. Good catch. It should be ready now.

peetucket (approved):

PR #669, opened 2021-04-12
Aub

jacobthill:
## Why was this change made?

Refreshing AUB data, fixing some data mapping issues.

Local transform and load in dev

n/a


lib/macros/language_extraction.rb
```diff
@@ -24,13 +24,25 @@ module LanguageExtraction
...
+            lang_code = val.match?(/[ضصثقفغعهخحمنتالبيسشظطذدزرو]/) ? arabic_script_lang : default
+            ar_values << val if lang_code == arabic_script_lang
+            default_values << val if lang_code == default
+          end
+          if ar_values.present? && default_values.present?
+            accumulator.replace([{ language: arabic_script_lang, values: ar_values }])
+            accumulator << { language: default, values: default_values } if default_values.present?
+          elsif ar_values.present? && default_values.empty?
+            accumulator.replace([{ language: arabic_script_lang, values: ar_values }])
+          elsif ar_values.empty? && default_values.present?
```
peetucket, line 43: There is a rubocop complaint about this method, and I think it could be simplified a bit since at this point in the if statement they are both empty or `default_values` exists ... so I think this branch could be simplified to `elsif default_values.present?`

This may still not satisfy rubocop.  At which point you could either try to split it into two methods, calling one from the other, or punt and just add an exception to rubocop around just this method like so:

```
# rubocop:disable Metrics/CyclomaticComplexity
    def arabic_script_lang_or_default(arabic_script_lang, default)
...
...
...
   end
# rubocop:enable Metrics/CyclomaticComplexity
```

You should be able to verify rubocop is happy before pushing:

`bundle exec rubocop`

peetucket (changes requested):
Added a note about a method that is causing a rubocop complaint and failing the build.

peetucket:
> @peetucket this fixes the rubocop errors but breaks the test. I can't figure out why?

I don't think the logic is quite the same with the refactored if statements if you follow all the possibilities. It's nice that the tests caught the difference, that shows they are working as expected.

I don't think it was that confusing the way it was before.  I would suggest disable both of the complaining cops for just that method and put it back the way it was before.  I've already tested this locally and the tests pass and no rubocop complaints when I disable the two offending cops.  I'll put a commit up.

peetucket:
Note: i just disabled the codeclimate maintenance checks going forward (and approved the one complaint it had about this method).  We use rubocop in this project and it is run automatically with the build on CircleCI, so having codeclimate also do similar checks is redundant.  This is the same pattern we have in other projects.  Codeclimate will still check test coverage though.

peetucket (approved):

peetucket:
Approved pending passing circleCI test.  Note: you may want to do consider doing a "squash and merge" commit unless you really want all of those back and forth commits in there.

PR #689, opened 2021-05-27
add new date macros, remove cruft

jacobthill:
## Why was this change made?

Several macros that do the same thing were abstracted into the `csv_or_json_date_range` macro which takes parameters instead of hard coding the date field paths in the macro. Old macros were removed.

n/a


ndushay (approved):

ndushay:
Approved; merge at will.

PR #697, opened 2021-06-10
yale babylon mapping

jacobthill:
## Why was this change made?

new dataset harvested

n/a

### DIFF
--- a/lib/translation_maps/getty_aat_materials_to_ar.yaml
+++ b/lib/translation_maps/getty_aat_materials_to_ar.yaml
@@ -1,0 +1,48 @@
+# Getty AAT materials in Arabic
+agate (chalcedony):
+bark (plant material):
+basalt (basic igneous rock):
+bone (material):
+bronze (metal):
+calcite:
+carnelian:
+ceramic (material):
+chalcedony:
+clay:
+copper (metal):
+dye:
+faience (composite material):
+feldspar:
+fiber:
+frit:
+glass (material):
+gold (metal):
+hematite (mineral):
+hide (collagenous material):
+iron (metal):
+jasper:
+lacquer (material):
+lapis lazuli (rock):
+lead (metal):
+limestone:
+marble (rock):
+magnetite:
+obsidian:
+onyx marble:
+paint (coating):
+paper (fiber product):
+pitch (tar):
+plant material:
+plaster (composite coating):
+plastic (material):
+quartz crystal:
+quartz (mineral):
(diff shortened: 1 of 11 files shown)


lib/translation_maps/getty_aat_materials_to_ar.yaml
```diff
@@ -0,0 +1,48 @@
+# Getty AAT materials in Arabic
```
aaron-collier, line 1: None of these have actual values, is that expected?

lib/translation_maps/periodo.yaml
```diff
@@ -0,0 +1,12365 @@
+'"Halaf" Chalcolithic':
```
aaron-collier, line 1: A few questions/thoughts about this config (not necessarily blockers, but to keep in mind perhaps for the next work cycle)

1.) It looks like all of the URIs are prefixed the same, will this always be true?
2.) This is a really long ~13K line config. Changes are likely to be challenging to maintain, we may consider getting this into a lookup table instead that could be more manageable.
3.) Some of these have several entries, have you tested the output to ensure the output data is formatted as expected?

lib/translation_maps/getty_aat_materials_to_ar.yaml
jacobthill: Yes, I just need to get the English values in so that someone at QNL can translate them. In the past I have managed translated via email or Transifex but I want to see if they can work directly out of the yaml files and submit PRs. That would simplify things.

lib/translation_maps/periodo.yaml
jacobthill: 1.) yes, I assumed the uris will always be prefixed the same way. I could remove the prefix and put it back in via the config. I guess that would save a lot of space.
2.) This file is generated automatically. PeriodO published the uris and lables and I fetch uris when labels match. I will periodically re-harvest the PeriodO data and re-run the script, then replace this file with the new output. I'm open to the lookup table if that makes more sense though, just noting that I won't be manually editing this file.
3.) I've manually inspected the output on several instances but haven't written any tests. It should just be an array of values that can be enforced via dry validation.

aaron-collier: That sounds good. It may be an area for later refactoring, but sounds like your current harvesting of it makes sense and we can discuss further as part of the work cycle.

aaron-collier (approved):

PR #698, opened 2021-06-10
Met

jacobthill:
## Why was this change made?

revise Met mapping

n/a


lib/macros/title_extraction.rb
```diff
@@ -39,18 +39,23 @@ def xpath_title_or_desc(xpath_title, xpath_desc)
...
         title = rec.xpath(xpath_title, NS).map(&:text).first
         other = rec.xpath(xpath_other, NS).map(&:text).first
-        if title.present?
-          if other.present?
-            acc.replace(["#{title} #{truncate(other)}"])
-          elsif other.present?
-            acc.replace([truncate(other)])
-          end
-        end
+        acc.replace(["#{title} #{truncate(other)}"]) if title.present? && other.present?
```
aaron-collier, line 47: Since both of these rely on `other.present?` I think it would be preferable to do something like:

```
if other.present?
  title.present? ? acc.replace(["#{title} #{truncate(other)}"]) : acc.replace([truncate(other)])
end
```

lib/macros/title_extraction.rb
```diff
@@ -39,18 +39,23 @@ def xpath_title_or_desc(xpath_title, xpath_desc)
...
+        acc.replace([truncate(other)]) if other.present? && title.empty?
+      end
+    end
+
+    # Extract a json title and truncated second field or, if no title in record, extract truncated second field.
+    def json_title_plus(json_title, json_other)
+      lambda do |rec, acc|
+        title = rec[json_title]
+        other = rec[json_other]
+        acc.replace(["#{title} #{truncate(other)}"]) if title.present? && other.present?
```
aaron-collier, line 57: Same as above here.

traject_configs/met_config.rb
```diff
@@ -7,7 +7,10 @@
 require 'macros/dlme'
 require 'macros/each_record'
 require 'macros/met'
+require 'macros/path_to_file'
+require 'macros/string_helper'
```
aaron-collier, line 11: Are these 2 requires necessary? I'm not seeing them used (though could just be missing them)

aaron-collier (changes requested):
Just a few comments/questions.

lib/macros/title_extraction.rb
jacobthill: Thanks, that a new syntax for me.

traject_configs/met_config.rb
jacobthill: `squish` is from `string_helper`. I know there is a rails method that does the same thing but I paired with John Martin to figure out how to use it and neither of us could figure out why it wasn't available. So we did this. I was missing the path_to_file but fixing in a follow up commit.

aaron-collier (approved):

PR #737, opened 2021-07-07
[WIP] Add test to validate temporal_from_contributor is translated to Arabic

### LINKED ISSUE #729: Add test to validate temporal_from_contributor is translated to Arabic
Add specs to ensure that each value in `temporal_from_contributor` exists as a key in `temporal_ar_from_en`.

See https://github.com/sul-dlss/dlme-transform/pull/718 as an example.

jmartin-sul:
## Why was this change made?

closes #729

unit tests

~~i followed the pattern in #718 as suggested in the ticket.  not sure if the test failures are because i did that incorrectly, or if the the thing being tested is actually failing validation appropriately.~~

only change is additional unit tests

### DIFF
--- a/spec/lib/translation_maps/temporal_ar_from_en_spec.rb
+++ b/spec/lib/translation_maps/temporal_ar_from_en_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+RSpec.describe 'temporal_ar_from_en translation map' do # rubocop:disable RSpec/DescribeClass this tests config consistency, not a class
+  let(:translation) do
+    indexer = Traject::Indexer.new
+    this = self
+    indexer.configure do
+      to_field 'cho_temporal', literal(this.key), translation_map('temporal_ar_from_en')
+    end
+    indexer.map_record(nil)['cho_temporal']
+  end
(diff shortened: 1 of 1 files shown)


spec/lib/translation_maps/temporal_ar_from_en_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+RSpec.describe 'temporal_ar_from_en translation map' do
+  let(:translation) do
+    indexer = Traject::Indexer.new
+    this = self
```
jmartin-sul, line 8: i was also trying to think of a slightly more descriptive name for `this`...

jmartin-sul:
paired w/ @aaron-collier after standup, and it seems like the tests in this PR are failing appropriately.  will `xit` this new test, and file a follow up ticket to un-skip it when we have the translations that will make it pass.

jmartin-sul:
might be superseded by another PR i'm working on that uses a shared example to cut down on boilerplate

jmartin-sul:
likely superseded by https://github.com/sul-dlss/dlme-transform/pull/739

PR #749, opened 2021-07-13
Refactor name extraction to introduce role mapping

### LINKED ISSUE #647: Revise extract_name macro in Mods module
As the DLME data manager I need to extract names and roles from MODS records. The current `extract_name` macro takes a role as argument, extracts a name if it has been assigned to that role, and appends the role in parenthesis after the name. e.g. 

```
to_field 'cho_contributor', extract_name('//*/mods:name[1][mods:role/mods:roleTerm/', role: 'scribe.'), arabic_script_lang_or_default('und-Arab', 'und-Latn')
```
However, it isn't compatible with the `arabic_script_lang_or_default` macro which checks for Arabic script and, if found, assigns the first parameter to `lang` else assigns the second parameter. 

If the extracted value is in English the `extract_name` macro works fine. If it is in any other language it will not work. e.g. if the extracted value is in Arabic, it will result in `'und-Arab': ['جزائري، محمد بن عمر بن إبراهيم، (scribe.)’]` – an Arabic value with an English role.

Requirements: 
- [x] The `extract_name` macro will be compatible with the `arabic_script_lang_or_default` macro so that if the extracted value is in Arabic script, a translation map is called that will take the role in English and fetch the Arabic translation. 
- [ ] The join for role in Arabic should be ` أو`; the join for exclude in Arabic should be ` و`
- [ ] A test should catch cases where an English role and an Arabic value falsely occur in the intermediate record.
- [x] Need method for translating English roles into Arabic; should we use a translation map? If so, it is okay to create one and grab test values from Google translate. The values can be professionally translated later (just tag @jacobthill in the ticket to alert him to this need).

The `harvard/islamic-heritage-project` data `cho_provenance` and `cho_contributor` fields have sufficient test data.

aaron-collier:
## Why was this change made?

Fixes #647 

This adds the translation of a role passed (and found) in the name parts of an extracted name into Arabic. Improves testing for existing macros used here. Removes empty `()` strings when the role is excluded from the name matching.

Improved unit tests.


spec/lib/traject/macros/mods_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+          </mods:name>
+        </mods>
+      XML
+    end
+    let(:ng_rec) { Nokogiri::XML.parse(record) }
+
+    context 'when contributor role matches the search' do
+      before do
+        indexer.instance_eval do
+          to_field 'cho_contributor', extract_name('//*/mods:name[1][mods:role/mods:roleTerm/', role: 'copyist.') # , strip, arabic_script_lang_or_default('und-Arab', 'und-Latn')
```
jmartin-sul, line 40: cruft?  or example of what would be there in more realistic usage?  something else?

jmartin-sul (approved):
a couple questions, and one very small nitpick to consider, but nothing i'd hold up merging over 👍

spec/lib/traject/macros/language_extraction_spec.rb
```diff
@@ -19,5 +19,14 @@
         expect(callable.call(nil, [extracted_string])).to eq([{ language: 'ar-Arab', values: [extracted_string] }])
       end
     end
+
+    context 'when extracted string contains Arabic characters and an english role' do
+      it 'returns the correct script value arabic_script_lang' do
+        extracted_string = 'الولايات المتحدة الامريكيه (Copyist)'
+        translated_string = 'الولايات المتحدة الامريكيه (الناسخ)'
```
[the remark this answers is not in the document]
aaron-collier: I'm not sure if it's traject or just right-to-left rendering magic. I don't have any code positioning the role, so this test validates that it is rendered right-to-left when appropriate.

spec/lib/traject/macros/mods_spec.rb
aaron-collier: cruft. good catch, thanks!

PR #760, opened 2021-07-21
 Remove generic cho_has_type value when specific value present

### LINKED ISSUE #701: Remove generic cho_has_type value when specific value present
As the DLME data manager I need to normalize cho_has_type values during transform. These values are mapped to a controlled vocabulary from terms in the contributor metadata using translation maps. Occasionally, the contributor data will contain multiple terms found in the translation map–one precise and one vague or too broad (e.g. the terms `weapons` & `silver` might both be used to describe an axe containing silver; `weapons` is mapped to the specific category `Armaments` and `silver` is mapped to a generic category `Other Objects`). When multiple values are found in the cho_has_type field and one of them is specific, the generic values should be removed.

generic_values = ['Other Image', 'Other Object', 'Other Text', 'Other Video', Other etc.] 

- [x] during transform, check length of cho_has_type array; if two or more objects, check each for string pattern 'Other *' . If all match string pattern, pass. If one does not match string pattern (the specific value), delete all values that do match string pattern, leaving only the specific value/s.
- [x] should be called from within the `normalize_has_type` macro so that all configs don't require updating or, if this isn't possible, will need to update all configs.

jmartin-sul:
## Why was this change made?

closes #701

also, some opportunistic touchups in the second commit (discovered the unused `let` vars while looking for testing patterns i could copy).

unit tests

n/a

### DIFF
--- a/lib/macros/normalize_type.rb
+++ b/lib/macros/normalize_type.rb
@@ -29,6 +29,10 @@ module Macros
                                Traject::TranslationMap.new(spec)
                              end.reduce(:merge)
         translation_map.translate_array!(accumulator)
+
+        # if there are type values more specific than "[O|o]ther *", keep only those more specific values
+        acc_minus_generic_types = accumulator.reject { |str| str.start_with?(/[O|o]ther /) }
+        accumulator.replace(acc_minus_generic_types) unless acc_minus_generic_types.empty?
       end
     end
   end
--- a/lib/macros/normalize_type.rb
+++ b/lib/macros/normalize_type.rb
@@ -21,11 +21,13 @@ module Macros
     # @example
     #   normalize_language => lambda { ... }
     # @return [Proc] a proc that traject can call for each record
-    def normalize_has_type
+    def normalize_has_type # rubocop:disable Metrics/MethodLength
       lambda do |_record, accumulator|
         accumulator.map!(&:downcase)
         translation_map = %w[has_type_from_contributor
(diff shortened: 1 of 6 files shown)


lib/macros/normalize_type.rb
```diff
@@ -21,14 +21,20 @@ def normalize_edm_type
...
     # @return [Proc] a proc that traject can call for each record
-    def normalize_has_type
+    def normalize_has_type # rubocop:disable Metrics/MethodLength
       lambda do |_record, accumulator|
         accumulator.map!(&:downcase)
         translation_map = %w[has_type_from_contributor
-                             has_type_from_fr has_type_from_lausanne has_type_from_tr].map do |spec|
+                             has_type_from_fr
+                             has_type_from_lausanne
+                             has_type_from_tr].map do |spec|
```
jmartin-sul, line 30: i found this slightly more readable than having one thing on one line and 3 things on the next, but i am happy to back this out if the old layout is preferable to others

aaron-collier, line 30: :+1:

aaron-collier: Yes! This was actually a change I was expecting during refactoring, 👏 happy to see it!

aaron-collier (approved):

lib/macros/normalize_type.rb
jmartin-sul: 😁

Issue #675, opened 2021-04-29
Require hash keys in IR to not equal 'none'


jacobthill:
As the DLME data manager, I need tests to ensure that no hashes in the intermediate representation are assigned `none` keys. Most metadata fields are assigned language keys; the default if not specified is ‘none’. No language key should be ‘none’. Since we are using BCP47 codes we can, at least, determine the script of the language. 

Requirements:
- [ ] Fails when `none` found, stop processing records; the transform process will stop and an error message with problematic field, values, and dlme_source_file will be reported to the terminal.
- [ ] Fails, only in debug mode, if field found with ‘*-Arab’ that includes Latin script; the transform process will stop and an error message with problematic field, values, and dlme_source_file will be reported to the terminal.
- [ ] Fails, only in debug mode, if field found with ‘*-Latn’ that includes Arabic script; the transform process will stop and an error message with problematic field, values, and dlme_source_file will be reported to the terminal.

jmartin-sul:
after discussing w/ @aaron-collier and @jacobthill in planning meeting on tues morning, we decided that we would just raise an error and allow the transform to fail fast when we run into a situation where the IR key would fallback to `'none'`.  however, when playing with that implementation yesterday, i discovered that we'd run into this situation frequently, even just within the aims data set, and so was worried about committing that change to `main` without first fixing the data and/or the transform configs (so as to prevent `'none'` from appearing as a key in the IR hashes).

discussed w/ @jacobthill just now, and we decided to do the following:
- [ ] to start, instead of raising an error, we'll log a message anytime we'd fall back to a hash key of `'none'`.
- [ ]  i'll run that code over the entirety of the dlme-metadata data set, collecting the output
- [ ]  i'll grep for the log message and give that output to jacob
- [ ]  he'll fix the records/configs as needed
- [ ]  once the data and/or configs are fixed up so that we don't have anymore of this error, we'll modify the code to raise an error in this situation, and then we'll fail fast and fix immediately if/when this problem gets introduced again in the future (by new data and/or configs).

jacob confirms that the hash keys always represent language (which i was unsure about, since i'm still relatively new to this project).

jmartin-sul:
[...]

PR #762, opened 2021-07-23
EachRecord.convert_to_language_hash: log an error when there's no language key

jmartin-sul:
also, small refactor to DRY up error message copypasta using a string template

## Why was this change made?

add logging to determine how frequently our transformation output is missing language.

ref #675

existing unit tests for regressions, run transform locally over entirety of dlme-metadata and send output to log file, look for instances of the new log message

n/a


jmartin-sul:
here's a grep for the new log message that catches missing language when running the transform script:

[missing_language_key_errors.log](https://github.com/sul-dlss/dlme-transform/files/6866657/missing_language_key_errors.log)

this is the transform output from `exe/transform -w` run across the entirety of [dlme-metadata](https://github.com/sul-dlss/dlme-metadata), filtered for the error message added in this PR.

looks like there are currently 610 occurrences of this issue.

jmartin-sul:
rebased on main after #766 was merged, but getting a new failure, described here:  https://github.com/sul-dlss/dlme-transform/issues/675#issuecomment-887914002

jmartin-sul:
rebased on main after #770 was merged, new error described here: https://github.com/sul-dlss/dlme-transform/issues/675#issuecomment-888643003

jmartin-sul:
will rebase and take out of draft once #776 is merged

aaron-collier (approved):

PR #763, opened 2021-07-26
add missing langauge keys

jacobthill:
## Why was this change made?

A new test to catch missing language keys was written. This PR adds keys to fields that are missing them so that PR can be merged.

local transform

n/a


jmartin-sul (approved):

jmartin-sul:
ref #675

PR #771, opened 2021-07-28
don't return null when no preview

jacobthill:
## Why was this change made?

at least one AUB record did not have a thumbnail and was failing type validation

local transform

n/a

### DIFF
--- a/traject_configs/aub_poha_config.rb
+++ b/traject_configs/aub_poha_config.rb
@@ -107,18 +107,22 @@ to_field 'agg_is_shown_at' do |_record, accumulator, context|
     'wr_id' => [extract_poha('/*/dc:identifierURI'), strip]
   )
 end
-to_field 'agg_preview' do |_record, accumulator, context|
-  accumulator << transform_values(
-    context,
-    'wr_dc_rights' => [literal('Available under a Creative Commons Attribution-Noncommercial-NoDerivatives '\
-                               '4.0 International License. Anyone is free to download and share works under '\
-                               'this license as long as they give credit for the original creation, the '\
-                               'shared work is not changed and not used for commercial purposes. '\
-                               'Attribution should be given to "AUB University Libraries." e.g. "Campus '\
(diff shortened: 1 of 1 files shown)


traject_configs/aub_poha_config.rb
```diff
@@ -107,18 +107,22 @@
...
+      'wr_dc_rights' => [literal('Available under a Creative Commons Attribution-Noncommercial-NoDerivatives '\
+                           '4.0 International License. Anyone is free to download and share works under '\
+                           'this license as long as they give credit for the original creation, the '\
+                           'shared work is not changed and not used for commercial purposes. '\
+                           'Attribution should be given to "AUB University Libraries." e.g. "Campus '\
+                           '1967" by AUB University Libraries is licensed under CC BY-NC-ND 4.0')],
+      'wr_edm_rights' => [literal('CC BY-ND: https://creativecommons.org/licenses/by-nd/4.0/')],
+      'wr_id' => [extract_poha('/*/dc:thumbnail')]
+    )
+  end
```
jmartin-sul, line 125: mostly asking for my own traject-related edification:  is it correct to say that the old version would apply the `transform_values` call to every record, pushing the result onto `accumulator`, and that the new version examines every record, applying the `transform_values` call to the record and pushing the result onto `accumulator, but skipping the record entirely if a thumbnail `xpath` isn't present?

wondering because this is the first time i've seen the `each_record` call in the middle of a bunch of `to_field` configuration calls.

jmartin-sul (approved):
one question since i haven't seen this exact construct before, but 👍 since it's tested and i _think_ i understand what's going on 🙂

jmartin-sul:
on second thought, refraining from merging till @aaron-collier can take a look.  i got hesitant since i don't totally grok the `to_field` in an `each_record` loop thing.

traject_configs/aub_poha_config.rb
jacobthill: Yes, this is correct. It will leave agg_preview out of the IR if a thumbnail xpath is not found.

aaron-collier (approved):

PR #772, opened 2021-07-28
Handling HTML tag in values

### LINKED ISSUE #679: Check all fields in the IR for html tags
As the DLME data manager, I need to ensure that records are not loaded with html tags embedded in the text values of any field. These tags can clash with default styles.

Requirements:
- [ ] Fails, only in debug mode, if any field contains html tags; the transform process will stop and an error message with problematic field, values, and dlme_source_file will be reported to the terminal.
- [x] When not in debug mode, does not fail; html tags are removed.

jermnelson:
## Why was this change made?
FIXES #679 

Adds unit test

n/a

## TODO
1.  [x]  Fail and log to screen in debug mode

### DIFF
--- a/lib/macros/each_record.rb
+++ b/lib/macros/each_record.rb
@@ -29,6 +29,11 @@ module Macros
             case value
             when Hash
               sub_values = value[:values].reject(&:nil?).reject(&:empty?)
+              html_cleaned = html_check(sub_values)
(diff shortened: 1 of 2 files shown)


jacobthill (commented):
I'll let a developer review this, but I  ran it on the AIMS set and can confirm it works.

aaron-collier (approved):

PR #776, opened 2021-07-29
Sundry config fixes

jmartin-sul:
## Why was this change made?

this splits out the traject config and macro changes @jacobthill made on top of the first two commits of #762.  since i didn't have any hand in those follow-on changes, and since they are in separate parts of the code changes i made in my first two commits, i can review them separately, and hopefully get this PR merged on its own.

CI, running transform locally on laptop

lots of traject configs


lib/macros/language_extraction.rb
```diff
@@ -50,6 +50,37 @@ def arabic_script_lang_or_default(arabic_script_lang, default)
...
+            val = translate_role_to_ar(val)
+            lang_code = val.match?(/[תשרקץצףפעסןנםמלךכיטחזוהדגבא]/) ? hebrew_script_lang : default
+            he_values << val if lang_code == hebrew_script_lang
+            default_values << val if lang_code == default
+          end
+        end
+        if he_values.present? && default_values.present?
+          accumulator.replace([{ language: hebrew_script_lang, values: he_values }])
+          accumulator << { language: default, values: default_values } if default_values.present?
+        elsif he_values.present? && default_values.empty?
```
jmartin-sul, line 75: i think this conditional could also be simplified: getting to this part implies that at least one of `he_values` or `default_values` is not present (since both have to be present to satisfy the `if`).  so this could just be `elsif he_values.present?` (because if `he_values` is present and the prior condition is false, `default_values` isn't present).

lib/macros/language_extraction.rb
```diff
@@ -50,6 +50,37 @@ def arabic_script_lang_or_default(arabic_script_lang, default)
...
+          accumulator.replace([{ language: hebrew_script_lang, values: he_values }])
+        elsif he_values.empty? && default_values.present?
```
jmartin-sul, line 77: by the same reasoning i think you could drop the `he_values.empty?` here, and just have `elsif default_values.present?`.

lib/macros/language_extraction.rb
```diff
@@ -50,6 +50,37 @@ def arabic_script_lang_or_default(arabic_script_lang, default)
...
+          accumulator.each do |val|
+            val = translate_role_to_ar(val)
+            lang_code = val.match?(/[תשרקץצףפעסןנםמלךכיטחזוהדגבא]/) ? hebrew_script_lang : default
+            he_values << val if lang_code == hebrew_script_lang
+            default_values << val if lang_code == default
+          end
+        end
+        if he_values.present? && default_values.present?
+          accumulator.replace([{ language: hebrew_script_lang, values: he_values }])
+          accumulator << { language: default, values: default_values } if default_values.present?
```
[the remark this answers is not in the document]
jmartin-sul: could you add a test that where `he_values.present? && default_values.present?` is true?

jmartin-sul (commented):
a couple suggestions for simplifying the logic in the new macro slightly, and i'd lean slightly towards testing the small part of the code that seems responsible for the coverage drop.  but otherwise, looks fine to me, thanks for all the testing and updating!

also @jacobthill i'm leaving this as a comment -- i went request changes, but this is technically my PR, even though the work and the commits are all yours, since i did the branch wrangling.

jmartin-sul (commented):
:tada: nice, looks good!  thanks for the touchups, @jacobthill!

one thing:  looks like a little bit of detritus got committed, maybe errant copy/paste on the CLI?  see image...

Screen Shot 2021-08-04 at 2 18 08 PM

aaron-collier (approved):

PR #787, opened 2021-08-11
add translations

jacobthill:
## Why was this change made?

- Add new Arabic translations
- Un-comment translation map tests


jacobthill:
@aaron-collier I tried to rebase this but I think I added some changes that weren't supposed to be added. I don't want to make it worse. Can you fix this please?

lib/translation_maps/edm_type_from_has_type.yaml
```diff
@@ -1,49 +1,53 @@
...
+Postcards: Image
+Posters: Image
+Recreational Artifacts: Object
+Reference Books: Text
+Sculptures: Object
+Structures: Object
+Seals: Object
+Serials: Text
+Stelae: Object
+Rablets: Object
```
jmartin-sul, line 46: `Rablets` -> `Tablets`?

jmartin-sul: ah, i see `Tablets` is in the list too, so maybe this translation is for a typo in the source data?

spec/lib/translation_maps/validate_contributor_translation_maps_spec.rb
```diff
@@ -25,29 +25,16 @@
 end
 
 RSpec.describe 'contributor translation maps are valid' do # rubocop:disable RSpec/DescribeClass this tests config consistency, not a class
-  # TODO: un-skip once we have the translations that will make it pass. see https://github.com/sul-dlss/dlme-transform/issues/738
```
jmartin-sul, line 28: should this TODO have stayed since the `it_behaves_like` below is still commented out?

jmartin-sul (changes requested):
a couple questions about changes to the translation maps, not sure if typos in translation maps or if typos upstream being accounted for by translation maps.  and one question about a removed TODO that seemed like maybe it should've stayed.

the "request changes" is more about the TODO, since i wasn't sure whether i'd noticed an actual problem with the translations.

spec/lib/translation_maps/validate_contributor_translation_maps_spec.rb
aaron-collier: Just noticed I missed that one. commit incoming.

lib/translation_maps/edm_type_from_has_type.yaml
aaron-collier: Ah - I totally missed `Rablets` and added `Tablets` - good catch.

spec/lib/translation_maps/validate_contributor_translation_maps_spec.rb
aaron-collier: I was supposed to also uncomment the test on line 28 (whoops).

jmartin-sul (approved):

PR #796, opened 2021-08-23
Simplify deployment instructions

aaron-collier:
## Why was this change made?

As the current work flow currently only utilizes dlme-transform in the development environment and dlme-transform is run as an ECS task and not a service, the image is pulled when the task is invoked. This makes the development deployment steps unnecessary and complicates stage and prod where this is rarely used.


jcoyne (approved):
🎉 🙌 🔥

PR #803, opened 2021-09-08
initial commit

jacobthill:
## Why was this change made?

We are building a collections registry and need a stable method for linking to the records that comprise a collection in DLME.
- Adds `dlme_collection` field and necessary validation
- Adds new translation maps (English and Arabic) and test to ensure that `dlme_collection` value is unique
- Updates configs

Local transform and rspec

Application profile is updated with new field: https://github.com/sul-dlss/dlme/pull/1328


jacobthill:
I'm not sure why deleting all of those tmp files is showing up here. I expected that to be ignored.

aaron-collier (approved):

PR #862, opened 2022-02-14
Remove Honeybadger from dlme-transform

aaron-collier:
## Why was this change made?

The original intention with including honeybadger in dlme-transform was to allow for alertable errors when detecting date or url issues that get past validation. This has proven to be too challenging during local development when most of these errors are originally detected anyway.

For now, we'd like to remove honeybadger to unblock local development and will in the future take another, more robust addition of honeybadger when alerting during automated transformation is more desirable.


aaron-collier:
@jacobthill I think the coverage drop here is fine, it's likely because this removes  a few honeybadger specific test, so the math changed a bit I suspect.

jacobthill (approved):

PR #867, opened 2022-02-23
add IFPO collection

jacobthill:
## Why was this change made?

Adds new collection and necessary translation map changes,

local transform and deploy to dev

n/a

### DIFF
--- a/config/metadata_mapping.json
+++ b/config/metadata_mapping.json
@@ -61,12 +61,12 @@
   },
   {
     "trajects": [
-      "aub_aco_config_csv.rb"
+      "aub_aco_config.rb"
     ],
     "paths": [
       "aub/aco"
     ],
-    "extension": ".csv",
+    "extension": ".xml",
     "settings": {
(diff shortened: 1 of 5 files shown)


config/metadata_mapping.json
```diff
@@ -61,12 +61,12 @@
...
   {
     "trajects": [
-      "aub_aco_config_csv.rb"
+      "aub_aco_config.rb"
     ],
     "paths": [
       "aub/aco"
     ],
-    "extension": ".csv",
+    "extension": ".xml",
```
aaron-collier, line 69: Shouldn't these stay csv? or will these be harvested outside of airflow?

aaron-collier (changes requested):

config/metadata_mapping.json
jacobthill: Good catch, these shouldn't be in here. I'll pull them out.

aaron-collier (approved):

PR #871, opened 2022-03-09
update aub settings

jacobthill:
## Why was this change made?

AUB POHA is manually shared and still needs to be mapped from xml

local transform

n/a

### DIFF
--- a/config/metadata_mapping.json
+++ b/config/metadata_mapping.json
@@ -106,7 +106,7 @@
     "paths": [
       "aub/poha"
     ],
-    "extension": ".csv",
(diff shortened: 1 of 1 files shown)


config/metadata_mapping.json
```diff
@@ -106,7 +106,7 @@
     "paths": [
       "aub/poha"
     ],
-    "extension": ".csv",
+    "extension": ".xml",
```
aaron-collier, line 109: Are we going back to XML for this collection?

aaron-collier (changes requested):

PR #900, opened 2022-07-28
add csv config for texas tech

jacobthill:
## Why was this change made?

Adding csv config

Local transform

n/a


aaron-collier:
@jacobthill small rubocop error should be quick to fix.

aaron-collier (approved):

PR #902, opened 2022-08-01
Remove agg_preview from penn due to no thumbnail provided

aaron-collier:
## Why was this change made?

@jacobthill in my airflow testing, I found that this transform fails due to the missing thumbnail column. If this is incorrect, please let me know so we can pair on it.


jacobthill (commented):
We don't want to do this. We need a post harvest step to fetch the thumbnails. I opened a ticket https://github.com/sul-dlss/dlme-airflow/issues/158

jacobthill:
@aaron-collier closing this, since we want to solve the missing preview problem in a post harvest task.

PR #903, opened 2022-08-03
Ssm

jacobthill:
## Why was this change made?

Transforms airflow csv output

local transform

n/a

### DIFF
--- a/config/metadata_mapping.json
+++ b/config/metadata_mapping.json
@@ -1590,12 +1590,12 @@
   },
   {
     "trajects": [
-      "ssm_abidindino_xml_config.rb"
+      "ssm_abidindino_config_csv.rb"
     ],
     "paths": [
-      "sakip-sabanci/abidindino-xml"
+      "sakip-sabanci/testing/abidindino"
     ],
-    "extension": ".xml",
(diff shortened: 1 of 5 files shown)


config/metadata_mapping.json
```diff
@@ -1610,15 +1610,52 @@
...
     "trajects": [
-      "ssm_iiif_config.rb"
+      "ssm_emirgan_config_csv.rb"
     ],
     "paths": [
-      "sakip-sabanci/abidindino-iiif",
-      "sakip-sabanci/emirgan",
-      "sakip-sabanci/kitap-vehat",
-      "sakip-sabanci/resim-klksyn"
+      "sakip-sabanci/testing/emirgan"
```
aaron-collier, line 1616: should this still be `.../testing/...`?

aaron-collier (changes requested):
the paths all includes `.../testing/...` should that be removed?

config/metadata_mapping.json
jacobthill: nope, good catch

aaron-collier (approved):

PR #911, opened 2022-09-01
draft auc config

jacobthill:
## Why was this change made?

Trying to map AUC from the csv airflow output. The data model is somewhat simpler now and hoping to reduce 22 configs into one slightly more complex config. To do so we need a more intelligent prepend macro.

Local testing, currently hung up on writing the tests for the new `intelligent_prepend` macro

n/a


spec/lib/traject/macros/prepend_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  let(:instance) { klass.new }
+
+  describe '#intelligent_prepend' do
+    context 'with en language code' do
+      let(:output) { [language: 'en', values: [+'value']] } # outputs nil
+
+      it 'prepends the English prepend string' do
+        callable = instance.intelligent_prepend('prepend string: ', 'السلسلة السابقة: ')
+        expect(callable.call(nil, output)).to eq([['prepend string: value']])
```
jacobthill, line 20: @aaron-collier I've added double brackets to all of these tests because that is what the output is. I'm not sure that's correct. Its passing but I assumed we wanted single brackets. I don't know why we would get/want a nested array. Any thoughts on this?

aaron-collier (approved):

PR #929, opened 2022-10-04
add more traject changes

jacobthill:


config/metadata_mapping.json
```diff
@@ -57,30 +57,14 @@
...
-    "trajects": [
-      "aub_aladab.rb"
-    ],
-    "paths": [
-      "aub/aladab"
+      "aub/aco",
+      "aub/aladab",
+      "aub/postcards",
+      "aub/posters",
+      "aub/travelbooks"
```
aaron-collier, line 67: I think you can just put `aub` here as it will recursively go down the tree.

aaron-collier (approved):
Approving with one comment that is non blocking.

PR #937, opened 2022-10-11
[HOLD] Remove XML dependency from qnl macros

### LINKED ISSUE #936: Refactor QNL Macros to use CSV data
The current QNL macros are still based on XML data, but we are now transforming csv based data.

https://github.com/sul-dlss/dlme-transform/blob/main/lib/macros/qnl.rb

The parts of this macro that are being used need to be refactored to utilize csv (i.e. replace xpath calls)

aaron-collier:
## Why was this change made?

Removes unused parts of the QNL macro, and fixes the used `generate_qnl_iiif_id` macro to work with the CSV data.

Closes #936 

Please hold until @jacobthill has been able to review the output

### DIFF
--- a/lib/macros/qnl.rb
+++ b/lib/macros/qnl.rb
@@ -3,78 +3,11 @@
 module Macros
   # Macros for extracting OAI values from Nokogiri documents
   module QNL
-    NS = {
-      oai: 'http://www.openarchives.org/OAI/2.0/',
-      mods: 'http://www.loc.gov/mods/v3'
-    }.freeze
-    private_constant :NS
-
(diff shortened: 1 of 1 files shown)


jacobthill:
I think we can accomplish everything we need without any macros. The only value the macros seemed to provide is to avoid duplicating long expaths in the traject config (which we don't do anymore now that we are mapping from csv). I will open another ticket to remove all of these unused macros completely. I think this will make the code easier to maintain and more readable.

aaron-collier:
@jacobthill in that case, I think this should be reopened and updated to remove the final macro if it can be, or a follow on PR. As this is removing unused macros.

PR #941, opened 2022-10-12
Avoid outputting nil when parsing csv array values

### LINKED ISSUE #939: parse_csv returns Nil in some cases
In some cases the `parse_csv` macro returns `Nil` which clashes with any traject methods that follow e.g. if `gsub` is called after. This is one of the reasons Harvard IHP is failing. 

`to_field 'cho_publisher', column('originInfo_publisher'), parse_csv...` is returning `Nil` and `arabic_script_lang_or_default`. I have confirmed this by commenting out that line and running the transform against the whole collection. It passes locally once I add some missing language values to our translation maps. 

To reproduce, you'll have to harvest all of the IHP set and run the config locally.

aaron-collier:
## Why was this change made?

Fixes #939 

Since often the array from the CSV becomes nested, we need to compact the array.

### DIFF
--- a/lib/macros/csv.rb
+++ b/lib/macros/csv.rb
@@ -20,7 +20,7 @@ module Macros
     # Determine if the value in the given column contains an array of values
     # return either the original value or the array value parsed into elements
     def parse_csv
-      lambda do |_row, accumulator, context|
+      lambda do |_row, accumulator|
         return [] if accumulator.empty?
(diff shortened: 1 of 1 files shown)


jacobthill (changes requested):
It looks like all of the tests need to be updates to take 3 arguments instead of 2.

jacobthill (approved):

PR #965, opened 2023-01-17
Bod

jacobthill:
## Why was this change made?

Add Bodleian collections back in and updates mapping, translation maps, etc. They were removed temporarily to give the data partner time to review.

Local transform and load in dev

n/a


edsu (approved):

edsu (changes requested):
It sounds like this is mostly re-enabling things that were once enabled, but I think maybe a unit test needs adjusting?

jacobthill:
> 

@edsu it should be good now

PR #985, opened 2023-03-15
Revise AUB

jacobthill:
## Why was this change made?

Fixes some issues in AUB transform

local transform and load in stage

n/a

### DIFF
--- a/lib/macros/path_to_file.rb
+++ b/lib/macros/path_to_file.rb
@@ -9,7 +9,7 @@ module Macros
     # @return [Proc] a proc that traject can call for each record
     def path_to_file
       lambda do |_, accumulator, context|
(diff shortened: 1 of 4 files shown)


lib/macros/path_to_file.rb
```diff
@@ -9,7 +9,7 @@ module PathToFile
     # @return [Proc] a proc that traject can call for each record
     def path_to_file
       lambda do |_, accumulator, context|
-        accumulator << File.join(Pathname(context.input_name))
+        accumulator << File.join(Pathname(context.input_name)).gsub('/opt/airflow/working', '')
```
edsu, line 12: Was this needed to get it working in your dev environment?

edsu (approved):

PR #986, opened 2023-03-15
Auc

jacobthill:
## Why was this change made?

Adds new AUC collections

local transform and loaded in stage

n/a


lib/macros/path_to_file.rb
```diff
@@ -9,7 +9,7 @@ module PathToFile
     # @return [Proc] a proc that traject can call for each record
     def path_to_file
       lambda do |_, accumulator, context|
-        accumulator << File.join(Pathname(context.input_name))
+        accumulator << File.join(Pathname(context.input_name)).gsub('/opt/airflow/working', '')
```
aaron-collier, line 12: This surprises me, but good to catch.

aaron-collier (approved):

PR #1048, opened 2023-11-29
fix data mapping errors

jacobthill:
## Why was this change made?

To fix some mapping errors

local transform

n/a

### DIFF
--- a/config/metadata_mapping.json
+++ b/config/metadata_mapping.json
@@ -1062,26 +1062,6 @@
   },
   {
     "trajects": [
-      "walters.rb"
-    ],
-    "paths": [
-      "walters"
-    ],
-    "extension": ".json",
-    "settings": {
-      "agg_provider": "The Walters Art Museum",
-      "agg_provider_country": "United States of America",
-      "agg_data_provider": "The Walters Art Museum",
-      "agg_data_provider_country": "United States of America",
-      "agg_provider_ar": "متحف والتر للفنون",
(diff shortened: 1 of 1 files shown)


config/metadata_mapping.json
```diff
@@ -1157,7 +1157,27 @@
...
+    "settings": {
+      "agg_provider": "The Walters Art Museum",
+      "agg_provider_country": "United States of America",
+      "agg_data_provider": "The Walters Art Museum",
+      "agg_data_provider_country": "United States of America",
+      "agg_provider_ar": "متحف والتر للفنون",
+      "agg_provider_country_ar": "الولايات المتحدة الامريكيه",
+      "agg_data_provider_ar": "متحف والتر للفنون",
+      "agg_data_provider_country_ar": "الولايات المتحدة الامريكيه",
+      "inst_id": "walters"
```
edsu, line 1180: did you meant to change this from "yale" to "walters"?

edsu (approved):
It looked like you copy and pasted an existing entry to the end, and then modified the existing entry to add a new one? It might be easier to read the diff if you edited the new entry, instead of modifying the old one?

I left a question for you to consider, but will leave it up to you to decide if it's an issue or not before merging.

config/metadata_mapping.json
jacobthill: I did change the "yale" id to "ucla" which is probably why it looks like that

PR #1067, opened 2024-02-27
fix walters mapping for new data

jacobthill:
## Why was this change made?

We are getting Walters data in a new way.

Local transform.

n/a

### DIFF
--- a/lib/contracts/cho.rb
+++ b/lib/contracts/cho.rb
@@ -111,7 +111,9 @@ module Contracts
     rule(:agg_provider_country, &required_language_specific_rule)
     rule(:agg_is_shown_at, &web_resource_validation_rule)
     rule(:agg_is_shown_by, &web_resource_validation_rule)
-    rule(:agg_preview, &web_resource_validation_rule)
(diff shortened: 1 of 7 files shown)


lib/contracts/cho.rb
```diff
@@ -111,7 +111,9 @@ def self.optional_language_specific_rule
     rule(:agg_provider_country, &required_language_specific_rule)
     rule(:agg_is_shown_at, &web_resource_validation_rule)
     rule(:agg_is_shown_by, &web_resource_validation_rule)
-    rule(:agg_preview, &web_resource_validation_rule)
+    # Calling web_resource_validation_rule on agg_preview results in a validation
+    # error for objects without thumbnail ids. We don't want this to happen.
+    # rule(:agg_preview, &web_resource_validation_rule)
```
jacobthill, line 116: Ideally, we should review this `web_resource_validation_rule`. I don't think we want it to raise an error on null values since cardinality is managed with `optional|required`

aaron-collier (approved):

PR #1073, opened 2024-03-15
Guard against nils in the accumulator when extracting language

aaron-collier:
## Why was this change made?

@jacobthill this avoids the nil issue from the test record you shared. Feel free to merge or close and just update your PR (this is against your branch)


aaron-collier:
@jacobthill I think the error here is related specifically to Manchester. The final 3 records they are providing are all blank. I will see if I can work out a way to avoid that on the harvest side - but I think that error can be ignored in this PR.

PR #1095, opened 2024-05-23
openiti revision

jacobthill:
## Why was this change made?

This maps in the description template text from a config maintained by the data provider.

local transform

n/a


lib/macros/openiti.rb
```diff
@@ -0,0 +1,31 @@
...
+        description = config['obj_descr_en'] if script == 'lat'
+        description = config['obj_descr_ar'] if script == 'ar'
+        values = {}
+        values[:author_ar] = record.fetch('author_ar')
+        values[:author_lat] = record.fetch('author_lat')
+        values[:text_url] = record.fetch('text_url')
+        values[:one2all_data_url] = record.fetch('one2all_data_url')
+        values[:one2all_stats_url] = record.fetch('one2all_stats_url')
+        values[:one2all_vis_url] = record.fetch('one2all_vis_url')
+        values[:pairwise_data_url] = record.fetch('one2all_data_url')
```
aaron-collier, line 22: I think this should be `record.fetch('pairwise_data_url')` unless we are purposefully duplicating `one2all_data_url` into `pairwise_data_url`.

jacobthill: Good eye

spec/lib/traject/macros/openiti_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'macros/openiti'
+require 'rspec'
+require 'webmock/rspec'
+require 'yaml'
+
+RSpec.describe Macros::OpenITI do # rubocop:disable RSpec/FilePath
```
jacobthill, line 8: This is raising two errors:

`RSpec/SpecFilePathFormat: Spec path should end with macros/open_iti*_spec.rb.`

and 

`RSpec/FilePath: Spec path should end with macros/open_iti*_spec.rb.`

I can't figure out how to disable both of them but it seems to be a rubocop error and not an actual error with my file name. https://github.com/rubocop/rubocop/issues/9505

traject_configs/openiti.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'dlme_json_resource_writer'
 require 'dlme_debug_writer'
+require 'httparty'
```
aaron-collier, line 5: What is `httparty` being used for?

jacobthill: To fetch the config file they provide from github. In theory, they will update this config periodically and we want to grab their updates whenever we run the DAG

aaron-collier: Gotcha, I don't think we need it, but I don't think that should block this - it's take a bit to get it working. I'll take a look later and ticket if we should refactor.

aaron-collier (approved):

PR #1102, opened 2024-06-13
Pin ffi to a supported version

aaron-collier:
## Why was this change made?

Pin ffi to avoid platform mismatch.


jacobthill (approved):
@lwrubel @edsu we were just talking about this

PR #1114, opened 2024-07-03
Move AUC to json, add validation, adjust macros

jacobthill:
## Why was this change made?

AUC was still using csv data and needed to switch to json. The json records contain nil values which causes, `strip`, `gsub`, `prepend`, etc. to fail. This adds new versions of those macros that guard against that. We are also seeing some errors with the values that get into `cho_edm_type` and `cho_has_type` so this treats those values as a controlled vocabulary and raises an error when a value not in the list is found. I'm planning a follow up PR to add nil-safe versions of all of the rest of the macros in https://github.com/traject/traject/blob/main/lib/traject/macros/transformation.rb. But this will require moving some things around and updating all of the configs that use them. Also planning another PR to remove all of the xml and csv macros that we no longer need.

Local transform, and new tests.

n/a

### DIFF
--- a/lib/macros/field_extraction.rb
+++ b/lib/macros/field_extraction.rb
@@ -13,12 +13,12 @@ module Macros
     }.freeze
     private_constant :NS
 
-    # Extracts value from the first column if available, or the second column if the first is not available.
-    def column_or_other_column(column, second_column) # rubocop:disable Metrics/AbcSize
-      lambda do |row, accumulator, _context|
-        return if row[column].to_s.empty? && row[second_column].to_s.empty?
+    # Extracts value from the field if available, or the default if the field is empty.
+    def extract_field_or_defualt(field, default) # rubocop:disable Metrics/AbcSize
+      lambda do |record, accumulator, _context|
+        return if record[field].to_s.empty? && record[default].to_s.empty?
 
-        result = row[column].to_s.empty? ? Array(row[second_column].to_s) : Array(row[column].to_s)
+        result = record[field].to_s.empty? ? Array(record[default].to_s) : Array(record[field].to_s)
         accumulator.concat(result)
       end
     end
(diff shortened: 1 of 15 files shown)


lib/macros/field_extraction.rb
```diff
@@ -13,6 +13,16 @@ module FieldExtraction
     }.freeze
     private_constant :NS
 
+    # Extracts value from the field if available, or the default if the field is empty.
+    def extract_field_or_defualt(field, default) # rubocop:disable Metrics/AbcSize
+      lambda do |record, accumulator, _context|
+        return if record[field].to_s.empty? && record[default].to_s.empty?
```
aaron-collier, line 19: I don't think you need to be doing the `to_s` before the `empty?` - something like `return unless record[field].present?` is probably more appropriate.

lib/macros/field_extraction.rb
```diff
@@ -13,6 +13,16 @@ module FieldExtraction
...
+
+        result = record[field].to_s.empty? ? Array(record[default].to_s) : Array(record[field].to_s)
```
aaron-collier, line 21: as as above.

lib/macros/prepend.rb
```diff
@@ -10,6 +10,7 @@ module Prepend
     def intelligent_prepend(prepend_string_en, prepend_string_translation)
       lambda do |_record, accumulator|
         return if accumulator.empty?
+        return if accumulator.first.nil?
```
aaron-collier, line 13: Is there a specific reason to check the first item for nil? Would this indicate the accumulator is nil or is there something special about the first item.

aaron-collier (changes requested):
A few comments to take into consideration.

aaron-collier (approved):

PR #1139, opened 2024-10-11
Move most remaining collections to transform from json

jacobthill:
## Why was this change made?

We've noted some issues with transforming from csv files. At times traject fails but in some cases it passes with minor formatting changes to the text. It seems safer and cleaner to simply transform all collections from json.

Local transform, rspec


jacobthill:
I believe all new macros have test coverage. The drop in coverage seems to be from deleting unneeded macros and their tests.

PR #1171, opened 2025-01-15
AUC collections have a depper path than previously, get proper value …

### LINKED ISSUE #1170: AUC transform task is resulting in `agg_data_provider_collection` not found error
AUC `transform` task is resulting in an error related to `agg_data_provider_collection` not being found. This works locally though so I don't know how to resolve this.

- [ ] Try to transform locally to confirm that it works locally ```docker run --rm -e SKIP_FETCH_DATA=true \
                -v $(pwd)/.:/opt/traject \
                -v $(pwd)/../dlme-airflow/working:/opt/airflow/working \
                -v $(pwd)/output:/opt/traject/output \
                suldlss/dlme-transform:latest \
                auc/general/abdullah_freres \
                -w```
- [ ] If it does work locally there might be an discrepancy between the local paths and those on airflow dev since we build the collection name from the file path: `path_to_file, dlme_split('/'), at_index(3), dlme_gsub('_', '-'), dlme_prepend('auc-')`. Resolve the issue
- [ ] Make sure an error in the `transform` task fails the task instead of simply logging it and passing the task.

aaron-collier:
…for lookup

Fixes #1170

### DIFF
--- a/traject_configs/auc.rb
+++ b/traject_configs/auc.rb
@@ -50,9 +50,9 @@ end
 to_field 'transform_version', version
 to_field 'transform_timestamp', timestamp
 
-to_field 'agg_data_provider_collection', path_to_file, dlme_split('/'), at_index(3), dlme_gsub('_', '-'), dlme_prepend('auc-'), translation_map('agg_collection_from_provider_id'), lang('en')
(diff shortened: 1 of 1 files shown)


jacobthill (commented):
This fails when I run it locally. That is the problem for me. There seems to be a different file path on airflow dev than I have locally. `at_index(3)` works locally. `at_index(5)` works on airflow dev. One command should work on both.

jacobthill (approved):

PR #1174, opened 2025-01-17
update Penn Museum mapping

jacobthill:
## Why was this change made?

Noticed missing fields when loading Penn Museum. The traject config was out of date.

local transform

n/a

### DIFF
--- a/lib/macros/prepend.rb
+++ b/lib/macros/prepend.rb
@@ -25,10 +25,10 @@ module Macros
 
         values = []
         accumulator.each do |val|
-          values << val
+          values << "#{prepend_string}#{val}" if val.present?
         end
 
-        accumulator.replace(["#{prepend_string}#{values.join(', ')}"]) if values.present?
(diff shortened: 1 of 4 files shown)


lib/macros/prepend.rb
```diff
@@ -25,10 +25,10 @@ def dlme_prepend(prepend_string)
 
         values = []
         accumulator.each do |val|
-          values << val
+          values << "#{prepend_string}#{val}" if val.present?
         end
 
-        accumulator.replace(["#{prepend_string}#{values.join(', ')}"]) if values.present?
```
jacobthill, line 31: I noticed while mapping this collection that when I had multiple harvested fields mapped to one dlme field, if any field was present all prepend strings showed up. So we had fields that looked like:

```
cho_description: "Some description text.",
Length: 
```
There was no value in the "length" field but because the "description" field had a value, the prepend string for "length" showed up in the IR.

aaron-collier (approved):

PR #1193, opened 2025-01-29
mapping changes, remove url validation, it chokes on nulls

jacobthill:
## Why was this change made?

New mapping changes and comments out url validation for wr_is_referenced_by because sometimes this is legitimately null and the validation won't pass null values. I left the comment so I can come back to it later but I don't want this to break the transform task.


lib/contracts/edm_web_resource.rb
```diff
@@ -3,15 +3,14 @@
...
   # See https://github.com/sul-dlss/dlme/blob/master/docs/application_profile.md#edmwebresource
   class EDMWebResource < Dry::Validation::Contract
-    # Basic URL regex.  Improve this for more robust validation.
-    URL_REGEX = %r{^(?:http(s)?://)}
-
     json do
       required(:wr_id).filled(:str?, format?: URL_REGEX)
       optional(:wr_format).array(:str?)
-      optional(:wr_is_referenced_by).maybe(:array).each(:str?, format?: URL_REGEX)
+      # optional(:wr_is_referenced_by).maybe(:array).each(:str?, format?: URL_REGEX) # Fails on null values
```
aaron-collier, line 13: Cruft? Can you remove this line or are we saving it for some reason?

jacobthill: In theory I would like to make sure that `wr_is_referenced_by` contains a valid url or [null] but its choking on null values at the moment. My plan was to come back to this are revise that so I left it.

aaron-collier (approved):

PR #1229, opened 2025-07-11
Adds Eastview serials and pattern for enriching metadata with Jaraid …

jacobthill:
…metadata

## Why was this change made?

- Adds the Eastview collection. This collection cannot be mapped through Airflow currently. The only way they have to share it is through FTP, so I manually grabbed the records.
- Adds a pattern for enriching DLME serial data with project Jara'id metadata. We are collaborating with MELA colleagues on a serials project. This is part of a proof of concept for the ETL portion.

- Local transform, and carefully examining records. Confirmed no duplicate ids and sample checked many urls.

Updated the application profile: https://github.com/sul-dlss/dlme/pull/1999


lib/macros/eastview.rb
```diff
@@ -0,0 +1,116 @@
...
+      return nil unless url_string_valid?(url_string, base_id, from_url_field, context)
+
+      extract_unique_part_from_url(url_string, base_id, from_url_field, context)
+    end
+
+    # Checks if the URL string is present and not empty.
+    # If not, logs a debug message and skips the record.
+    # @return [Boolean] true if valid, false otherwise.
+    def url_string_valid?(url_string, base_id, from_url_field, context)
+      if url_string.nil? || url_string.empty?
```
aaron-collier, line 46: I think `url_string.blank?` would cover both cases here.

lib/macros/eastview.rb
```diff
@@ -0,0 +1,116 @@
...
+      context.skip!("Malformed URL in field '#{from_url_field}': #{url_string}")
+      nil
+    end
+
+    # Validates the 'd' parameter from the parsed query parameters.
+    # Logs a debug message and skips the record if the 'd' parameter is missing or empty.
+    # @return [String, nil] The 'd' parameter value if valid, nil otherwise.
+    def validate_and_get_d_param(params, url_string, base_id, from_url_field, context)
+      unique_part = params['d']
+      if unique_part.nil? || unique_part.empty?
```
aaron-collier, line 82: Same as above, this could simply be:
```
unique_part.blank?
```

aaron-collier (approved):
Pending consideration of small changes.

PR #1277, opened 2026-03-25
Add Eastview records

jacobthill:
## Why was this change made?

Adds the Eastview collection which was shared via FTP

Local transform

n/a


jacobthill:
An important pattern change:

- I am trying to normalize serial data to support a collaborative project with MELA. Project Jara'id provides useful and clean metadata for a subset of Arabic serials. We are using there xml files combined with newly added translation maps to map Jara'id ids to other ids and then fetch structured metadata.

aaron-collier (approved):

